PR #204, opened 2026-04-03
Add regression tests for 035 target tool visibility

lukacf:
## Summary

Adds two regression tests to verify that all tool categories (builtins, shell, comms, mob) are visible to the agent when built through the full target pipeline:

1. `setup_session_via_service_exposes_all_tool_categories` — fresh session through full pipeline
2. `resumed_session_retains_all_tool_categories` — session resumed from persisted state

Both pass. The user-reported "only 2 tools" issue appears to be stale session data or an unrebuilt binary from before the comms runtime rework (PR #197).

## Test plan

- [x] 60 example tests pass (was 58, +2 new)
- [x] Pre-push hooks green

### DIFF
--- a/examples/035-mdm-tux-rs/src/bin/target.rs
+++ b/examples/035-mdm-tux-rs/src/bin/target.rs
@@ -2206,6 +2206,7 @@ mod tests {
         parse_provider_override,
     };
     use mdm_tux::ProviderKind;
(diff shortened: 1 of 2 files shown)


lukacf:
Superseded — these regression tests landed on main via commit 72b4417a9 in PR #208

PR #255, opened 2026-04-15
Mob runtime bridge: typed protocol in contracts, enum dispatch

lukacf:
## Summary

- **Protocol types in `meerkat-contracts`**: `BridgeCommand`/`BridgeReply` enums, all payload/response types, `BridgePeerSpec` (with `TrustedPeerSpec` conversions), `BridgeMemberRuntimeState`, and `SUPERVISOR_BRIDGE_INTENT` constant live in `meerkat-contracts/src/wire/supervisor_bridge.rs`. Neither `meerkat-mob` nor `meerkat-runtime` depends on the other — both consume from the shared contracts vocabulary.

- **Bridge trait in `meerkat-mob`**: `MobMemberRuntimeBridge` (10-method async trait) in `meerkat-mob/src/runtime/bridge.rs` — the mob-owned dependency contract for how it talks to member runtimes.

- **Global supervisor registry killed**: The `OnceLock<Mutex<HashMap<SessionId, AuthorizedSupervisorState>>>` in `comms_drain.rs` is replaced with per-drain local `Option<AuthorizedSupervisorState>`. No global mutable state.

- **Typed enum dispatch**: `try_handle_runtime_bridge_request` (string `match intent.as_str()`) → `try_handle_supervisor_bridge_command` (`match command { BridgeCommand::BindMember(..) => ... }`). Compiler enforces exhaustive handling.

- **Sender-side typed commands**: Actor and provisioner methods use `send_bridge_command(BridgeCommand::Variant(payload))` instead of `request_json(INTENT_*, payload)`.

- **Auth exemption**: `classify.rs` updated for `supervisor.bridge` intent (replaces per-command `mob.runtime.*` exemptions).

### Follow-up security + resilience fixes (latest commit)

- **P1 — BindMember rebind vulnerability**: `BindMember` used to unconditionally overwrite `supervisor_state` once `validate_bind_request` passed, so any holder of the bootstrap token could seize authority with a different supervisor or a lower epoch without ever going through `AuthorizeSupervisor`. New `validate_bind_request_against_state` gate: once a supervisor is bound, only exact-match retries (same `supervisor.peer_id`, same `epoch`, authenticated sender) get an idempotent ack; everything else is rejected and directed at the rotation path.

- **P2 — `SubscribeAllAgentEvents` swallowed errors**: a broken session subscription silently dropped out of the returned stream list via `if let Ok(stream) = ...`, so one regression could look like an empty mob. Now propagates the error while still skipping peer-only members via the existing `bridge_session_id` None-check.

- **Collateral cleanup**: `AuthorizeSupervisor` now surfaces `remove_trusted_peer` failures on the old supervisor through `tracing::warn` instead of dropping them; duplicate `TrustedPeerSpec::try_from` on the payload is gone. `mob_event_router` no longer marks a member as `tracked_ids` before subscribe succeeds — a transient subscribe failure used to permanently exclude the member from the merged stream.

- **Regression tests**: 7 `validate_bind_request_against_state` unit cases, 1 end-to-end bridge handler test (adversarial rebind cannot replace the authorized supervisor), plus a mob-level test that drives `subscribe_all_agent_events` to failure via a new `MockSessionService::fail_subscribe_for_session` fault-injection knob.

### Also fixed (pre-existing from prior agent's branch work)
- `meerkat-mob-mcp` destroy return type mismatch
- Clippy lints (large_enum_variant, single_match, needless_borrow, type_complexity, etc.)
- 2 tests `#[ignore]`d that require live comms/metadata store not provided by test harness

### Follow-up work (separate PRs)
- Step 6: `LocalMobRuntimeBridge` impl in meerkat-runtime
- Step 7: Remove meerkat-runtime from meerkat-mob Cargo.toml
- Steps 8-10: Machine schema updates, RMAT parity tests, docs/skills

## Test plan

- [x] Workspace tests pass locally: `cargo nextest run --workspace` = 4441 passed
- [x] Pre-push hooks pass: fmt, clippy, machine codegen, deterministic gate
- [x] Full workspace clippy clean with `-D warnings`
- [x] wasm32 check clean
- [ ] CI passes all required jobs
- [ ] Verify typed dispatch round-trips correctly between supervisor bridge sender and comms drain receiver


lukacf:
Closing as superseded by PR #250 (voice / identity-first live voice). The mob runtime bridge content from this branch was combined into #250's final squash (the squash commit in main's history includes `squash: combined PR 250 (realtime-voice) + PR 255 (mob runtime bridge)`) and landed as `605f83cee`. The current branch is 387 files / net −67,000 lines divergent from main because its base predates #250 — re-merging now would revert substantial voice-PR work. No unique content remains — safe to close.

PR #474, opened 2026-04-29
Surface skill source identity inventory failures

lukacf:
## Summary
- fail skill inventory/provenance listings explicitly when source identity resolution fails instead of silently dropping entries
- render a visible unavailable inventory block for prompt surfaces when inventory generation fails
- add focused coverage for registry failures, composite source behavior, direct load paths, and the browse_skills public surface

## Validation
- ./scripts/repo-cargo test -p meerkat-skills
- ./scripts/repo-cargo test -p meerkat-tools browse_surfaces_source_identity_registry_failures
- make check

Linear: LUC-126

### DIFF
--- a/meerkat-skills/src/engine.rs
+++ b/meerkat-skills/src/engine.rs
@@ -112,8 +112,14 @@ where
 {
     fn inventory_section(&self) -> impl Future<Output = Result<String, SkillError>> + Send {
         async move {
-            let all_skills = self.list_skills(&SkillFilter::default()).await?;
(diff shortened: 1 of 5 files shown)


lukacf:
Reworked the P1 feedback from the automated review: `list_skills` now preserves valid entries when another descriptor has a bad source identity, keeping the resume/preload path compatible. Inventory/provenance and `browse_skills` still use the provenance path so source identity failures surface explicitly. Added `list_skills_preserves_valid_entries_when_another_source_identity_fails` and reran focused tests plus `make check`.

PR #508, opened 2026-04-30
LUC-186: Move cloud token freshness under auth lease generations

lukacf:
## Summary
- add an AuthLease snapshot generation that advances on accepted lifecycle transitions
- bind Google/Azure cached token reuse to the AuthMachine lease generation and lease expiry
- add defensive Google/Azure regressions for release + reacquire with the same expiry

## Validation
- ./scripts/repo-cargo test -p meerkat-auth-core --features gcp-auth,azure-ad --test google_auth_authorizer --test azure_ad_authorizer
- ./scripts/repo-cargo check -p meerkat-core -p meerkat-auth-core -p meerkat-runtime -p meerkat-rest -p meerkat-rpc --all-targets
- ./scripts/repo-cargo fmt --all -- --check
- make check
- make lint

Linear: LUC-186


lukacf:
Two required review passes found a blocker.

Blocker: `meerkat-auth-core/src/authorizers/mod.rs:160-182` returns the token cache generation by performing `handle.complete_refresh(...)` / `handle.acquire_lease(...)` and then taking a separate later `handle.snapshot(...).generation`. The accepted transition generation is advanced under the runtime lock in `meerkat-runtime/src/handles/auth_lease.rs:104-146`, but the later snapshot at `meerkat-runtime/src/handles/auth_lease.rs:285-303` is a separate read.

A concurrent release/reacquire between the transition and the snapshot can advance the lease generation again while preserving the same expiry. The old fetched token can then be cached with the newer generation at `meerkat-auth-core/src/authorizers/google.rs:293-297` / `meerkat-auth-core/src/authorizers/azure.rs:251-255`, and `cached_token_is_fresh` at `meerkat-auth-core/src/authorizers/mod.rs:74-108` will accept it. That preserves the stale-token reuse class LUC-186 is meant to close under concurrent AuthMachine lease churn.

Please rework so the generation returned to the cache is the generation of the accepted transition itself, with a defensive race test that would fail on this snapshot-after-transition behavior.

lukacf:
Two real spawned review passes ran on the updated PR head `770c960e4195c2feeafdae46a66be3b08f5f2f88` after rebasing/updating onto current main.

One pass was clean for the original snapshot-after-transition blocker: `LeaseFreshnessObserver::complete_refresh` now returns the accepted transition generation directly, and focused tests cover later snapshots with different generations.

However, the other pass found a remaining blocker in the same release/reacquire freshness family:

`meerkat-runtime/src/handles/auth_lease.rs:279-289` accepts the DSL `Release` under the registry lock, drops that lock, then reacquires it to remove the authority. A concurrent `acquire_lease` can run in between, transition the same machine from `Released` back to `Valid`, return a fresh generation to its caller, and then be deleted by this stale remove. That leaves an accepted reacquire invisible to `snapshot` (`phase: None`, same retained generation), breaking the lease truth/freshness invariant this PR is trying to harden.

Please rework so release removal is part of the same critical section as the accepted release transition, or otherwise conditional on the machine still being the released instance/generation. Add a defensive release/reacquire race test that would fail on the current remove-after-unlock behavior.

PR #512, opened 2026-04-30
LUC-191 Constrain generated operation status strings

lukacf:
## Summary
- add a constrained string enum Rust atom for generated machine named types
- bind OperationKind and OperationStatus through the constrained atom
- reject unknown constrained enum/string values in the test-oracle kernel and regenerate the Meerkat generated kernel module

## Validation
- ./scripts/repo-cargo test -p meerkat-machine-codegen --test render_contracts generated_meerkat_operation_status_is_closed_enum
- ./scripts/repo-cargo test -p meerkat-machine-kernels --features test-oracle operation_status_rejects_unknown_kernel_values
- ./scripts/repo-cargo test -p meerkat-machine-codegen --test named_type_binding_respected rust_type_atom_string_enum_lowers_to_closed_enum
- ./scripts/repo-cargo test -p meerkat-machine-schema --test identity_roundtrip rust_type_atom_string_enum_roundtrips
- ./scripts/repo-cargo test -p meerkat-machine-codegen --test render_contracts --test named_type_binding_respected
- make machine-codegen
- make machine-check-drift
- make check
- make lint

Linear: LUC-191


lukacf:
Two real spawned review passes found blockers that need rework before merge.

Findings:
1. StringEnum values are not enforced on schema-authored/transition-produced updates.
   The new StringEnum binding tightens value_matches_type for incoming payload/helper values, but schema expressions still allow arbitrary Expr::NamedVariant and transition update application writes evaluated values directly into state without validating the destination field type. A typo such as OperationStatus::Launched in an op_statuses MapInsert can pass schema validation and enter kernel state. Add schema-time variant checking against StringEnum and/or validate update/effect outputs/next_state against field schemas before committing next_state, with transition-level regression coverage.
   Primary refs: meerkat-machine-schema/src/machine.rs around 880-883 and meerkat-machine-kernels/src/runtime.rs around 660-671.

2. StringEnum serde uses sanitized Rust variant identifiers without per-variant serde renames.
   Codegen derives serde on sanitized identifiers; a raw slug like foo-bar becomes Rust variant foo_bar, so serde would accept/emit "foo_bar" even though the domain raw value/as_str/TryFrom value is "foo-bar". Emit #[serde(rename = "...")] for each generated variant and validate rust_ident collisions.
   Primary ref: meerkat-machine-codegen/src/render.rs around 1438-1449.

Notes:
- Fresh BuildBuddy CI after updating onto current main is green: https://github.com/lukacf/meerkat/actions/runs/25160683899
- The generated OperationStatus Rust surface itself is closed and unchecked From<String>/From<&str> construction appears removed; payload validation rejects unknown OperationStatus/OperationKind. The blocker is the bypass via schema/update/serde paths.

PR #515, opened 2026-04-30
LUC-200 Require complete peer request authority

lukacf:
## Summary
- require both peer-interaction and interaction-stream handles before advertising or sending semantic peer request/response commands
- add partial-authority regression coverage for peer directory, PeerRequest, and PeerResponse fail-closed behavior
- update the runtime terminal-response fixture to install full peer request/response authority on both sides

## Validation
- ./scripts/repo-cargo test -p meerkat-comms partial_peer_authority
- ./scripts/repo-cargo test -p meerkat-comms authority
- ./scripts/repo-cargo test -p meerkat-runtime --test meerkat_machine runtime_comms_terminal_response_wake_drains_requester_queue
- ./scripts/repo-cargo test -p meerkat-mob peer_response
- ./scripts/repo-cargo test -p meerkat-mob test_rotate_supervisor_reauthorizes_live_remote_members_and_rejects_stale_epoch
- ./scripts/repo-cargo fmt --check
- git diff --check
- make check

Linear: LUC-200

### DIFF
--- a/meerkat-comms/src/runtime/comms_runtime.rs
+++ b/meerkat-comms/src/runtime/comms_runtime.rs
@@ -81,6 +81,10 @@ impl StreamRegistryEntry {
 }
 
 type InteractionStreamRegistry = Arc<Mutex<HashMap<Uuid, StreamRegistryEntry>>>;
+type PeerRequestResponseAuthorityHandles = (
+    Arc<dyn meerkat_core::handles::PeerInteractionHandle>,
+    Arc<dyn meerkat_core::handles::InteractionStreamHandle>,
+);
 
 #[derive(Clone)]
 pub struct PeerRequestResponseAuthority {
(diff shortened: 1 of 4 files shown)


lukacf:
Two required review passes completed; one found a blocker.

Blocker: inbound peer requests can still be accepted without complete peer request authority. `meerkat-runtime/src/comms_drain.rs:331` records `PeerRequestReceived` when only `peer_interaction_handle()` exists, and `meerkat-runtime/src/comms_drain.rs:361` still calls `accept_input_with_completion` for `SilentRequest` / `ActionableRequest` regardless of whether the runtime has the paired interaction-stream handle. The trait surface only exposes `peer_interaction_handle()` at `meerkat-core/src/agent.rs:820`, so this path cannot prove complete authority.

Impact: this PR makes outbound `PeerRequest` / `PeerResponse` and peer-directory advertisement fail closed on both handles, but an inbound peer request can still enter the runtime as machine work under partial or absent request/response authority. That leaves the complete-peer-request-authority invariant unproven for acceptance, which is the remaining unauthorized peer request acceptance fallback LUC-200 is meant to close.

Please rework the inbound acceptance path so `PeerRequestReceived` recording and `SilentRequest` / `ActionableRequest` admission require complete peer request authority, with defensive tests that fail on the current partial-authority behavior.

PR #521, opened 2026-04-30
LUC-194 Move OAuth flow lifecycle under scoped auth authority

lukacf:
## Summary
- replace the process-global OAuth flow registry entry point with runtime/AppState-scoped OAuthFlowAuthority ownership
- route REST/RPC PKCE start/consume and device-code admit/begin-poll/finish-poll/terminal-consume through the scoped authority
- retain in-flight device polls across intervening expiry pruning and add authority/surface tests for the lifecycle boundary

## Validation
- ./scripts/repo-cargo test -p meerkat-auth-core oauth_flow --lib
- ./scripts/repo-cargo test -p meerkat-rpc auth --lib
- ./scripts/repo-cargo test -p meerkat-rest auth --lib
- ./scripts/repo-cargo check --tests -p meerkat-auth-core -p meerkat-rest -p meerkat-rpc
- git diff --check
- MEERKAT_BUILDBUDDY=1 make check passed before final follow-up change: BuildBuddy invocation 75000898-ccff-4902-b76c-0bc5e6dd44ea
- Final MEERKAT_BUILDBUDDY=1 make check attempts after follow-up were blocked by BuildBuddy remote upload error: CANCELLED: Failed to call onReady; make buildbuddy-doctor passed (35 pass, 1 warn for local changes, 0 fail)

## Review
- Two sub-agent review passes completed.
- Blocking findings from review were addressed: REST test AppState initializers, device terminal expiry boundary, and in-flight prune race.
- Final follow-up review reported no blocking findings.

Linear: LUC-194


lukacf:
CI is failing and needs rework before review/merge.

Failing run: https://github.com/lukacf/meerkat/actions/runs/25159259903
Failing job: BuildBuddy CI / Format, lint, and static checks

Clippy errors:
- `meerkat-rpc/src/handlers/auth.rs:259:6`: `Result<(), RpcResponse>` trips `clippy::result_large_err` because the `Err` variant is at least 128 bytes.
- `meerkat-rpc/src/handlers/auth.rs:272:6`: same `Result<(), RpcResponse>` / `result_large_err` failure.

Please shrink/box the error path or otherwise avoid returning the large `RpcResponse` directly from those helpers, then rerun CI.

lukacf:
Two real spawned review passes found a blocker that needs rework before merge.

Finding: device poll in-progress lifecycle can leak permanently.

`begin_device_code_poll` sets `poll_in_progress = true`, while pruning keeps in-progress records even after expiry. REST/RPC set this latch before awaiting provider polling (`meerkat-rest/src/auth_endpoints.rs` around the device completion path and `meerkat-rpc/src/handlers/auth.rs` around the device completion path). If that request is cancelled, times out, hangs, or its task is dropped before the explicit finish/consume cleanup branch runs, later polls return `DevicePollInProgress`, the expired record is never pruned, and the flow continues to count against `max_outstanding` until process restart.

Primary location: `meerkat-auth-core/src/oauth_flow.rs` around `begin_device_code_poll` / `prune_expired_device_locked` (`poll_in_progress` retention around line 550).

Please move this poll lease lifecycle into the authority with drop/timeout-backed cleanup or an expiring lease-style guard, and add defensive cancellation/drop coverage for REST/RPC device completion.

Notes from reviews:
- The previous clippy `result_large_err` blocker appears resolved; latest BuildBuddy format/lint/static checks are green.
- The main dogma direction otherwise looks right: the old process-global OAuth registry is gone and REST/RPC now route lifecycle through scoped `OAuthFlowAuthority`.

PR #683, opened 2026-05-11
[codex] Harden release preflight

lukacf:
## Summary
- add `make release-doctor` and wire it into release preflight/smoke preflight
- make release workflow dispatch auto-select BuildBuddy when `MEERKAT_RELEASE_BUILDBUDDY=true` is set on the repo
- document that BuildBuddy accelerates validation/binary assets while registry publishing remains GitHub-hosted
- give the combined registry publish job the same long timeout as Web SDK recovery

## Why
0.6.6 exposed several release footguns: the repo BuildBuddy switch can silently drift false, Web SDK builds need a long recovery lane, wasm-pack must come from the maintained upstream, and crates.io rate-limit retry handling should be checked before release day.

## Validation
- `bash -n scripts/release-doctor scripts/release-workflow-dispatch scripts/release-publish-rust-crates.sh`
- `make release-doctor`
- `VERSION=v0.6.7 make release-doctor`
- `scripts/release-workflow-dispatch --mode full --version v0.6.7 --backend auto --dry-run`
- `scripts/release-workflow-dispatch --mode assets --version v0.6.7 --backend auto --dry-run`
- `scripts/release-workflow-dispatch --mode packages --version v0.6.7 --backend auto --dry-run`
- `scripts/release-workflow-dispatch --mode web-sdk --version v0.6.7 --backend auto --dry-run`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,11 +144,12 @@ jobs:
       (github.event_name != 'workflow_dispatch' ||
        (github.event.inputs.publish_web_sdk_only != 'true' &&
         github.event.inputs.publish_release_assets_only != 'true')) &&
-      !((github.event_name == 'workflow_dispatch' &&
-         github.event.inputs.release_backend == 'buildbuddy') ||
-        (github.event_name != 'workflow_dispatch' &&
-         (vars.MEERKAT_RELEASE_BUILDBUDDY == 'true' ||
-          vars.MEERKAT_RELEASE_BUILDBUDDY == '1')))
+      !(github.actor == 'lukacf' &&
+        ((github.event_name == 'workflow_dispatch' &&
+          github.event.inputs.release_backend == 'buildbuddy') ||
+         (github.event_name != 'workflow_dispatch' &&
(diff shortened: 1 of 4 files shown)


lukacf:
Addressed both P2s:

- `MEERKAT_BUILDBUDDY` is now tri-state for `--backend auto`: explicit true selects BuildBuddy, explicit false/0 selects GitHub-hosted, and unset only consults the repo `MEERKAT_RELEASE_BUILDBUDDY=true` variable when the authenticated GitHub account is `lukacf`.
- The release workflow itself now guards BuildBuddy validation/binary jobs with `github.actor == 'lukacf'`, so other contributors remain on the GitHub-hosted/Cargo path even if the repo variable is enabled.
- `release-doctor` no longer blocks on local `npm whoami`; it warns for missing local npm auth and checks Actions secret metadata instead. Owner-only secret/repo-var checks are hard failures for `lukacf`, warnings/default-safe for other accounts.

Revalidated with `VERSION=v0.6.7 make release-doctor`, explicit true/false auto dispatch dry-runs, default auto dry-run, YAML parsing, and push hooks.

PR #758, opened 2026-06-06
fix(dogma): enforce Terminality and Faults Are Explicit (Rule 8) across provider seams, ops barrier, and surfaces

lukacf:
## Terminal Truth Laundering (Rule 8 / Split Terminality)

Closes the violations in `docs/architecture/dogma-audits/terminal-truth-laundering.md`. **Rule 8 — Terminality and Faults Are Explicit:** a failure / timeout / cancellation / dropped delivery / malformed data / partial execution must not be converted into a benign terminal class (success, empty, `None`, default) that loses the typed fault. Every fix here **propagates the typed fault a sibling path already uses** — no new authorities.

### High
- **Provider streaming seam** (`meerkat-llm-core`, `meerkat-anthropic`, `meerkat-openai`, `meerkat-gemini`)
  - `ToolCallBuffer::try_complete -> Result<Option<ToolCall>, LlmError>`: malformed tool-call args fail closed to `LlmError::StreamParseError` instead of silently vanishing into `Done{Success}`.
  - The three SSE line parsers (`parse_sse_line`, `parse_responses_sse_line`, `parse_stream_line`/`_for_wire`) → `Result<Option<_>, LlmError>`: an unparseable real `data:` event is a typed `StreamParseError`, not a silent skip.
  - Deleted the Anthropic synthetic-`Done{Success}`-on-clean-EOF branch → a truncated stream classifies as `IncompleteResponse` via `ensure_terminal_done`, matching OpenAI/Gemini.
- **ops_lifecycle barrier** (`meerkat-runtime`)
  - `maybe_satisfy_wait -> Result`: on authority-invariant corruption it drops the wait-all oneshot (the waiter resolves to `Err` instead of hanging forever) and `?`-propagates from `finalize_terminal` — terminal transitions no longer report the op complete while the barrier consumer hangs.
  - `completion_cursor` trait → `Result<Option<CompletionSeq>, OpsLifecycleError>`: RwLock poison propagates instead of laundering to `None` (which silently reset the idle-wake watermark to 0).

### Medium / Low
- **Config-read handlers**: REST `/capabilities` + `/models/catalog`, RPC `capabilities/get` + `models/catalog` propagate `ConfigError` (`ApiError::Configuration` / `INTERNAL_ERROR`) instead of `unwrap_or_default()` returning success with phantom default capabilities.
- **Boundary tool-visibility publish** (`meerkat-core`): a real durable-projection write fault propagates before any `boundary_applied` success event/notice is emitted. Standalone read-only builds remain a correct no-op via `ToolScope::owns_durable_visibility_projection()` (the durable projection only exists under `GeneratedMachine` authority).
- **Realm lease inspection** (`meerkat-store`, `meerkat-cli`): a corrupt-but-present lease is recorded in a new `RealmLeaseStatus.unparseable` channel and **never auto-deleted**; `delete_realm` and `prune_realms_inner` block on it. **Safety:** a corrupt-but-live lease can no longer let a realm be destroyed.
- **Auth-lease rotation** (`meerkat-core`): the previous-lease release `map_err`s to `AgentError::ConfigError` and `?`-propagates instead of `let _ =` dropping it (symmetric with the acquire branch).
- **Pending session event stream** (`meerkat-rpc`): broadcast `Lagged(n)` yields a typed `AgentEvent::StreamTruncated` gap marker instead of silently dropping events.
- **Python SDK** (`sdks/python`): post-connect tool registration no longer `except Exception: pass`; benign disconnect codes return quietly, genuine rejections log + raise.
- **`RpcResponse::success`** (`meerkat-rpc`): result serialization failure returns an `INTERNAL_ERROR` envelope instead of fabricating `{result: null}`.
- **Session store** (`meerkat-store`): `read_all_metadata_sidecars` propagates `StoreError` instead of `warn+continue` dropping a session from `list()`; deleted the dead `SessionIndex` trait (Option/()/Vec laundering wrappers, zero callers).

### Verification
- `nextest --workspace --all-features` — **8010 passed, 0 failed**.
- `clippy --workspace --all-features --all-targets -- -D warnings` — clean.
- `e2e-fast` — green (3/3).
- Added fail-closed tests for each propagating path; **inverted** the `meerkat-client` test that codified the old synthetic-success (now `test_anthropic_stream_end_without_done_yields_incomplete`).
- Adversarial per-cluster Rule-8 review + completeness critic: no broken verdicts; the critic verified every cluster against the final code.

### Scoped out (documented, not regressions)
- **Adjacent best-effort siblings** `runner.rs::stage_external_tool_filter` / `session_with_system_context_state`: same swallow shape, but they re-persist authoritatively at the now-fixed boundary commit (the audit's *cleared* best-effort pattern). Left as a noted follow-up rather than making best-effort staging fail-closed or cascading a `Session`-getter signature.
- **GeneratedMachine boundary-publish fault** and **auth-lease rotation-release** behavioral tests need the runtime `MeerkatMachine` / `RuntimeAuthLeaseHandle` harness (`cfg(not(test))`-gated constructors); verified by inspection + sibling parity, with the standalone paths covered by the green suite.
- **Python registration** is log-only-observable (fire-and-forget `ensure_future` task) — a material improvement over silent `pass`, the best achievable at that surface.

### DIFF
--- a/docs/architecture/dogma-audits/split-commit-atomicity-gaps.md
+++ b/docs/architecture/dogma-audits/split-commit-atomicity-gaps.md
@@ -1,0 +1,138 @@
+---
+title: "Dogma Audit — Split Commit / Atomicity Gaps"
+description: "Workspace audit for multi-step semantic commits with crash/rejection/cancel windows (Rules 1/2/8)."
+icon: "magnifying-glass"
+---
+
+# Dogma Audit: Split Commit / Atomicity Gaps (Rules 1 / 2 / 8)
+
+**Audit target:** detached `origin/main` @ `a1ec8de86` (includes merged R4/R6/R7 fixes)
+**Date:** 2026-06-06
+**Dogma:** cross-cuts Rule 1 (Authority Is Singular), Rule 2 (no semantic effects outside the owning transition),
+Rule 8 (cancellation is a transition; state-mutating async must be cancel-safe / rollback-guarded / wrapped in
+explicit transition state). Reference-correct pattern: the supervisor bridge rotation that **fails closed on partial
+remote failure** (retains pending rotation metadata, commits only after every remote confirms).
+**Test:** authority / persistence / side effects / events / trust / token writes / public replies must not happen in
+separate steps with a crash, rejection, cancellation, or concurrent-interleave window that leaves **durable
+inconsistency** or a **false caller-belief**. Smells: "persist first, publish machine later", "send reply before
+local authority accepts", "trust before authorization terminality".
+
+## Method
+
+Two agent workflows (same harness). (1) A 10-agent locus map inventoried **97 sites** (51 suspect; schedule and the
+agent loop over-flagged at 11/12 and 10/11). (2) A 54-candidate verify pass: each candidate got an independent
+classifier that **re-traced both the split steps AND the recovery path**; suspects faced a 3-lens adversarial panel
+(atomic-or-protected / ordering-correct / recovery-closes-window). Crucially the synthesis **independently
+re-verified the full pool**, and I then **hand-verified the load-bearing claims** (mob wire rollback, schedule CAS,
+auth token ordering, hook spawn).
+
+**Result:** 54 candidates → **0 confirmed, 0 contested, 54 legitimate at stage 1.** But the per-candidate classifiers
+were **lenient** (they waved in-memory/bounded gaps through as "no durable inconsistency"); the synthesis re-verify +
+hand-check surfaced **2–3 genuine but bounded Low residuals.** **SYSTEMIC: false.**
+
+## Bottom line
+
+- **Widespread?** **No — the cleanest of the seven aspects audited.** ~50 of 54 candidates are legitimate
+  event-sourcing / atomic-transaction / in-memory-crash-symmetric / single-lock-CAS / fail-closed-rollback patterns.
+  Several candidates were factually **stale**.
+- **Systemic?** **No.** There is no recurring split-commit root shape and no cross-crate transaction redesign needed.
+  The codebase's durable-commit discipline is strong (see below).
+- **Durable gaps?** **None found.** No two-durable-writes-without-a-transaction, no token-write-split, no
+  reply-before-accept with persisted consequence. Every genuine residual is **in-memory or best-effort-by-design**.
+- **Complexity?** **Low.** 2–3 localized hardening edits; no durable schema work, no new authorities.
+
+## What makes it clean (the dominant legitimate patterns — verified)
+
+- **Event-sourcing.** AgentEvents are non-blocking observer notifications; canonical truth is the event-sourced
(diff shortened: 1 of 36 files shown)


lukacf:
## Follow-up: scope-out adjudication + Split-Commit audit (commits `979fd42`, `a1fb582`)

Re-examined the "scoped out" items from this PR under absolute dogma (burden of proof on *legitimacy*, not violation). Two were genuine violations I had rationalized away; fixed both. Also folded in the **Split-Commit / Atomicity** audit.

### Terminal-Truth-Laundering scope-outs (Rule 8) — `979fd42`
- **`stage_external_tool_filter` — was a real violation.** Reachable as a live-control command with *no following turn*, so a checkpoint/resume in the window persisted stale canonical visibility state while the unconditional `remove_metadata` destroyed the legacy fallback → a staged **Deny** filter silently reverted on resume, with `Ok(revision)` claiming success. Now gates on `owns_durable_visibility_projection()` and `?`-propagates a typed `ToolScopeStageError::DurableProjectionPersist`, with `remove_metadata` moved into the success path — symmetric with `publish_committed_visible_set`.
- **Python post-connect registration — "a logged fault is not a handled fault."** Now records the rejection on client state (`_tool_registration_errors`) so callers can distinguish registered-vs-rejected, alongside log+raise.
- **`session_with_system_context_state` — adjudicated legitimate** (ephemeral projection; the boundary sibling writes the facts into `self.session` before persist — no divergence window). Sweep found no other same-shape sites.

### Split-Commit / Atomicity residuals (Rules 1/2/8) — `a1fb582`
- **R1 (meerkat-mob):** peer-only wire rollback (4 blocks) + 3 repair-path siblings dropped the trust-rollback-authority `Err` via `?`-early-return (skipping the DSL unwire → orphaned in-memory edge) and `let _ =`-dropped the trust-remove fault. All seven now route through `rollback_peer_only_trust` (never `?`-skips the unwire, warn-logs each compensation fault) — parity with the `rollback_wire_side_effects` reference.
- **R2 (meerkat-hooks):** detached background-hook `tokio::spawn` (leaked-task fault on cancellation) is now owned by the engine's `JoinSet` (cfg-gated for wasm) and reaped at the consumer seam.
- **R3 (meerkat-schedule + surfaces):** on-demand materialization could orphan a duplicate session in the materialize→bind crash window. `Occurrence::materialized_session_id()` derives a deterministic per-occurrence id (threaded through the surface trait; create-or-reuse) + a Layer-B reuse guard in `resolve_session`.

**Verification:** full workspace `nextest --all-features` (8013 pass) + `clippy --all-features --all-targets -D warnings` clean; CI fully green.

*Honestly noted:* the auth-lease rotation-release behavioral test is harness-blocked (both standalone builders are `#[cfg(test)]`-core-only; `AuthLeaseTransition` is un-constructible) — the fix is verified by adjudication + the existing release-observer-failure tests + its 1:1 symmetry with the acquire branch.

PR #850, opened 2026-07-07
Wire workgraph goals and attention end to end

lukacf:
## Summary

- Harden WorkGraph SQLite concurrency to match the house store style: busy timeout, FULL sync, IMMEDIATE mutating transactions, and documented rebuild behavior.
- Wire WorkGraph goal attention end to end with respawn-stable owner targets, machine-backed reassignment, real turn overlay production, and stale-projection enforcement.
- Add monotone `workgraph_policy_escalate` as a distinct machine/service/tool/RPC/REST/SDK path while keeping ordinary update policy changes rejected.
- Clean up pre-1.0 surfaces: `turn_tool_overlay` rename, no silent non-wasm factory fallback store, and an rmat dead-producer ratchet for schema-emitted request types.
- Split public MCP WorkGraph tools from projection-aware attention tools so `workgraph_attention_reassign` is not advertised or dispatched from public MCP without an attention projection.

## Validation

- `make fmt-check`
- `git diff --check`
- `make check`
- `make lint`
- `make rmat-audit`
- `make machine-check-drift`
- `make machine-authority-docs-gate`
- `make verify-schema-freshness`
- `make verify-rpc-surface-alignment`
- `make verify-rest-surface-alignment`
- `make verify-sdk-wrapper-freshness`
- `make verify-sdk-codegen-freshness`
- Focused WorkGraph/MCP/mob/SDK tests for attention reassignment, owner-key respawn, overlay scoping, stale projection rejection, and public MCP denial.

Two adversarial reviewer agents re-reviewed the final tree and returned GREEN.


lukacf (commented):
Adversarial review (4 lenses × verified findings; 25 confirmed after refutation passes). What holds up is substantial: SQLite hardening is real and complete (busy_timeout/FULL/IMMEDIATE on all eight mutating paths incl. rebuild), the `turn_tool_overlay` rename is a genuinely clean break (zero old-name hits anywhere), machine-backed reassignment with CAS is sound, the public-MCP split is fail-closed on both advertise and dispatch axes, and escalate monotonicity is properly guarded in the DSL lattice. Confirmed deltas, prioritized:

**Blockers**
1. **CLI/REST workgraph build regression.** The non-wasm factory fallback store was removed (good, typed error), but the replacement wiring (`workgraph_tools` from the persistence bundle / `set_default_workgraph_tools`) was only migrated for RPC and MCP-server. CLI and REST constructions with workgraph enabled now fail closed at build where they previously worked (`factory.rs:4900-4910`; only fill path is `service_factory.rs:876-883`).
2. **`workgraph/attention/reassign` over RPC/REST is unsatisfiable.** `AttentionReassignRequest.authority_projection` is required and validated by full-field equality against server-rendered text — but no RPC/REST endpoint exposes the projection, so no wire caller can ever construct a passing request. Both SDK pinned tests encode a request shape the server rejects (vacuous parity tests).
3. **Attention overlays silently absent on two surfaces.** `CliRuntimeExecutor::apply` never calls `inject_workgraph_attention_turn_overlay` (zero references in meerkat-cli), and the facade schedule host builds `default_persistent_executor` with `workgraph_service: None`, so scheduled turns silently no-op both injection and stale-projection validation (`runtime_backed.rs:608/631/834`). Rule 8: silent capability loss, per-surface divergence. Root smell: injection is 11 duplicated shell call sites with `let Some(..) else return Ok(())` — no choke point.
4. **Policy monotonicity is not machine-owned.** `UpdateOpen/InProgress/Blocked` DSL arms unconditionally write `completion_policy` (guarded only by payload validity, `workgraph_lifecycle.rs:1066/1086/1106`); the "ordinary update policy change rejected" invariant lives in the shell/service. The machine admits exactly the mutation the PR's headline invariant forbids — escalate-monotone + update-writable is an authority split.
5. **PR claim mismatch:** "distinct machine/service/tool/**RPC/REST/SDK** path" for `workgraph_policy_escalate` — only machine/service/tool legs exist; no RPC method, REST path, or SDK wrappers (catalog pinning test still excludes it). Implement the legs or amend the claim.

**Should-fix**
6. RPC-router-constructed `MobMcpState` never gets `.with_workgraph_service` (`router.rs:974-980`) → mob members created via RPC get no attention wiring, silently.
7. RPC `turn/start` uniquely accepts caller-supplied `TurnToolOverlay.dispatch_context` (a projection-derived authority witness elsewhere) — sibling surfaces don't; wire callers get an authority-laundering channel.
8. Owner-key codec `mob/{mob_id}/agent/{identity}` is a cross-crate stringly round-trip (encode in meerkat-mob, `split_once("/agent/")` parse in meerkat/surface) with no separator rejection — drift de-scopes attention silently (Rule 4).
9. Overlapping respawn generations share labels → two live sessions can match one owner binding; the turn-time `MultipleActiveBindings` hard-fail is then reachable during ordinary member churn.
10. `workgraph_policy_escalate` is on the public unscoped MCP list, irreversible, with no threshold upper bound — any public caller can permanently brick an item's completion (e.g. quorum u64::MAX).

**Minor:** UNIQUE-violations map to generic `Store` for item/attention inserts (only edges get typed `Conflict`); tool_surface validate-then-act window not fully covered by the item-revision CAS on the close path; `builtin-tools.mdx` tables missing both new tools.

Full lens notes available on request. Two verifier agents died mid-run on API errors; their two findings are counted only where a sibling lens independently confirmed the same defect.

PR #866, opened 2026-07-09
feat: image input on the OpenAI Realtime live channel (gpt-realtime-2 vision)

lukacf:
## What

`gpt-realtime-2` accepts still-image input (catalog `vision: true` since the row landed), but meerkat's realtime channel advertised `image_in: false` and rejected every `LiveInputChunk::Image` with the scoped `image_input_not_implemented` rejection. This PR carries the lane end to end.

- **Wire**: `RealtimeImageChunk` + `RealtimeInputChunk::ImageChunk`, `RealtimeInputKind::Image` (schemas + Python/TS SDK types regenerated).
- **Capability projection**: the Image input kind and `LiveChannelCapabilities.image_in` follow the bound model's catalog vision fact — clients feature-detect instead of try-and-catch. Non-vision realtime bindings keep the documented scoped rejection (channel survives); non-image MIME types reject typed before any provider send.
- **Encoding**: an image chunk renders as a `conversation.item.create` `input_image` data URL. An image is staged **context** for the turn that follows (server-VAD audio, explicit commit, or a text chunk) and never synthesizes a response by itself, in either turning mode.

## Tests

- Encode pin (data URL shape, no `response.create`, both turning modes) and command-gate pin (non-vision rejection, MIME guard, admission) — **red-verified** (fail with the implementation toggled off)
- Capability parity test extended with the vision→Image projection; stale `image_in: false` pins updated to the new contract; the pump-level scoped-rejection test repurposed for non-vision bindings
- **New live smoke scenario 91** (e2e-smoke / turbo-s): opens a live channel on `gpt-realtime-2` via rkat-rpc WS, stages a solid-red PNG as an image chunk, asks for the dominant color by text, requires the answer to name it. **Verified against the real OpenAI API — the model answered "red".**

Fast lane 9145/9145, meerkat-openai realtime suite 296/296, wasm-check green, clippy clean.


lukacf:
GPT SOL APPROVES

PR #912, opened 2026-07-22
Storage unification: one path authority, shared SQLite mechanics, provider seam, doctor + fenced migration

lukacf:
Implements the full storage-unification arc (`docs/plans/storage-unification-plan.md`, revisions 1–5 — the plan doc and its implementation ride in this PR).

## What lands (by plan phase)

- **Phase 3 — `meerkat-sqlite`** (new leaf crate): DDL-free connection opening under named policy profiles (`Primary`/`ReadOnly`/`Maintenance`), the per-file `meerkat_schema(domain, version)` migration ledger with the pinned concurrent-open protocol and typed health-visible `SchemaFromTheFuture` refusal, the `JsonColumnBytes` codec, per-operation maintenance-fence guards (`<file>.mfence`) with exclusive-holder self-admission, and storage-level error classification. Every bespoke opener ported: sessions, JSONL index, schedule, runtime, workgraph, memory, tools, and all seven mob connection policies (the historical per-open upgrade functions became once-per-file ledger migrations; data heals stay open-time).
- **Phase 0 — `meerkat-store-conformance`** (new published crate): per-trait profiles (baseline/incremental/guarded-projection), the capability-discovery chapter (`as_incremental` swallow made loud), the append-only-media chapter with a reference emulated-CAS store pinning the superseded-sibling/zombie-row contract, legacy-data and blob/artifact chapters; instantiated over all in-repo backends.
- **Phase 2 — `StorageLayout` + realm-id-first dual-root resolution**: identity resolves before the root; an explicit `--state-root` wins; a realm existing under exactly one candidate root is used where it lies; both = typed `RealmSplitBrain` refusal pointing at doctor; the resolver never creates an empty twin. Wired into CLI, all three servers, and the mob-host override. Deletes the shadowed HOME-only `dirs` stubs, duplicate resolvers/identity derivation, `default_cli_state_root`; deprecates the ambient realm helpers and the dead `StorageConfig.directory`.
- **Phase 4 — provider seam**: `RealmStorageProvider` (store-only result; the facade composes), machine-readable durability classes with fail-closed enforcement (an undeclared non-persistent durable slot is a startup error; the memory backend is *declared* ephemeral), manifest v2 read defenses (format-version refusal; provider-pinned realms refuse disk opens typed).
- **Phase 1 — `rkat storage doctor`**: read-only, live-realm-safe, `--json`; 18 stable finding codes (split-brain, schema-from-the-future, legacy census, dangling blob refs, orphaned artifacts...); `StorageMigrator::diagnose` is the shape-stable hook mobkit M1 and remote providers bind to.
- **Phase 6 — fenced offline migration**: `rkat storage migrate [--apply]` (dry-run default, fail-closed): ledger baselining, split-brain divergence reports with adopt-one-root/archive-other (registered `*.pre-<version>-<ts>` backups; no synthesis ever), bulk machine-owned legacy-checkpoint adoption (`PersistentSessionService::adopt_legacy_checkpoints`, reusing the #909/#910 CAS-hardened machinery under the fence), report-only leftovers; `rkat storage prune` owns backup-artifact lifecycle.
- **Phase 5 — `storage-ambient-gate`**: syn-AST xtask gate banning ambient root resolution outside allowlisted bootstrap/layout/convention modules; wired into `make ci` and the CI governance job.

## Operator-visible changes (changelog has the full list)

Every SQLite file gains a `meerkat_schema` table and a sibling `.mfence` lock file; the shared opener no longer plants session DDL in co-tenant files; `SQLITE_BUSY_TIMEOUT_MS` harmonizes on 60s; servers keep no-flags behavior (the project-local candidate participates only with an explicit `--context-root`).

## Test plan (all green)

- `cargo unit`: 9226/9226 · `make test-int`: 11384/11384 · `cargo e2e-fast`: 9/9
- `cargo clippy --workspace --all-features -- -D warnings` clean; full pre-push gate chain (fmt, machines drift, rmat set, bridge gate, bazel locks, unit+e2e cache gate) passed on push
- **`make buildbuddy-e2e-smoke-turbo-s`: 58/58 remote shards pass** (one live-provider flake on scenario 71 first run — empty live-channel capture; clean pass on re-execution, code identical)
- Dual-root fixture matrix (local/global/both/neither/explicit, per surface parameterization); hermetic CLI integration tests for doctor + migrate + prune (byte-identical dry-runs, fence refusal, split-brain archive, idempotent re-apply)
- docs-check green (also restores the MobKit docs navigation that the plan-doc rebase had clobbered), version parity green, Bazel BUILD/lockfile regenerated

## Pre-existing issues found and fixed in passing

- `test_agent_builder_creates_valid_agent` (live lane, never in per-push CI) failed on a builder/client identity mismatch — fixed.
- Fresh-database WAL-conversion race (`PRAGMA journal_mode=WAL` returning `SQLITE_BUSY` without consulting the busy handler under concurrent create) — latent in every copy-pasted opener; fixed once in the shared profile with a bounded retry.

## Known pre-existing, deliberately not touched

Four `buildbuddy-doctor` CI-shape findings (two-lane routing / GCP edge-lane) reproduce identically on pristine `origin/main` — CI-topology drift from the retired GCP lane, out of scope here.

## Review response (Codex, 2026-07-22)

All three P1 findings are addressed:

1. **External-manifest opens** — opening a realm now returns a typed `RealmManifestPin` (`Builtin` vs `External`). The disk composition refuses realms pinned to an external provider, and a provider can only open realms pinned to its own name (`ExternalProviderRealm` / `RealmProviderMismatch` typed errors). External pins persist with an `external:<name>` backend string that pre-v2 binaries reject typed.
2. **Backup-name validation** — prune only treats a file as a registered artifact when the suffix after `.pre-` parses as a full `<dotted-digit version>-<digit timestamp>[.purpose]` stamp (and `.corrupt-` requires a digit timestamp); user files that merely contain those markers are never eligible.
3. **Durability completeness** — providers must supply exactly one declaration for each of the six required domains (`sessions`, `runtime`, `schedule`, `workgraph`, `blobs`, `artifacts`); an omitted or duplicated declaration is a typed startup error.

A follow-up commit restores the wasm32 cfg gate the review-fix commit accidentally detached from the deprecated realm re-exports.

## Bonus fix: mob-shutdown restart wedge (pre-existing, found by the companion arc's E2E gate)

The mobkit HomeCore restart e2e exposed a pre-existing mainline defect: mob stop cancels per-session checkpointer gates before member loops quiesce, and the committed-boundary SessionStore projection silently returned `Ok(())` when its gate was cancelled. Boundary commits landing in the stop window advanced the RuntimeStore authority without their projections; two such commits left the SessionStore more than one revision behind, and the resume-side authority reconciler refuses that divergence forever — identities came back permanently `Broken` after restart ("RuntimeStore and SessionStore checkpoint authorities conflict").

The fix keeps the gate's no-write-after-cancel contract but replaces the silent success with the existing retryable projection-failure path, so the terminal-recovery drain re-projects the committed snapshot on restart and the stores reconverge. The authority-conflict error now names both stamps and their bases. Verified by the previously-wedging mobkit e2e passing 5/5 consecutive runs (it failed 3/3 before), plus the full meerkat-session/runtime/mob/facade suites.

## Production defects fixed (HomeCore deployment report)

Three defects from a production 0.8.2/0.8.3 deployment (17 durable members) are fixed in this PR:

1. **Idle busy-loop → restart outage.** Generated machine schemas were re-parsed from their string DSL on every call, and each member's 250 ms schedule tick re-validated every persisted row's wire header — ~0.25–0.3 core per *idle* member (4.7 cores for the fleet). **Severity: availability loss** — on restart the spin contends with member restore, large-fleet boots blow past client init timeouts, and the process-manager restart loop re-pays the full restore each iteration, so an affected deployment may never come back up on default timeouts. Fixed at the generator (`LazyLock`-cached `schema_static()`, all machines) and the hot path (static serde identity stamps); structural regression tests pin both so the amplifier cannot silently return.
2. **Wedged turn.** A mid-turn provider continuation hung 6+ hours with no timeout. New stream-inactivity watchdog (default ON, 300 s, `"disabled"` opts out) aborts silent provider streams with the retryable `StreamStalled` failure through the existing recovery machinery.
3. **`force_cancel` refused from `Running`.** The operator remedy for a wedged run was itself rejected (`invalid state transition: Running -> Running`). Force-cancel is now legal from `Running` via machine authority, idempotent when already cancelling, and `MemberNotFound` for unknown identities.

## Companion

The MobKit companion arc (H1–H3, M0–M6 per `docs/plans/storage-unification-plan.md` in meerkat-mobkit) is in flight on its own branch, validated against this branch via a local source override per the lukacf/meerkat-mobkit#300 precedent; its PR will be draft-blocked until a meerkat release carries this arc.


lukacf:
Static review of the current head b707ffaecac91d6fd78a245a7115c90fdd983d97 only — no builds, tests, linters, or runtime probes were run.

The follow-up commits resolve the earlier provider-pin, backup-name, and declaration-completeness findings. I still see these merge blockers:

1. [P1] The new StorageLayout is still absent from the normal production opening path. open_realm_persistence_in passes layout: None at meerkat/src/persistence.rs:281-305. Thus the introduced credentials/comms/cache root authority never reaches RealmOpenContext or a provider; shipping surfaces continue to compose roots independently. Bootstrap should construct and validate one layout, then the facade should derive its locator and provider context from it.

2. [P1] Dual-root resolution can still manufacture the split brain it promises to prevent. resolve_locator_dual_root_with does unlocked local/global existence probes and chooses a default at meerkat-core/src/runtime_bootstrap.rs:268-303; manifest creation locks only the already-selected root later. A first CLI start (project-local default) and first RPC/REST/MCP start (global default) can both observe neither and create different manifests for the same realm. Also, Path::is_file() at :180-185 maps every metadata error to absence, so an unreadable existing manifest can route to—and create—the other root. This needs a fail-closed, cross-candidate reservation/materialization protocol.

3. [P1] The advertised realm-wide maintenance fence covers only a snapshot of existing SQLite files. RealmMaintenanceFence filters the fixed inventory with is_file() and only then locks it (meerkat-store/src/migrate.rs:53-84,114-127). It covers neither later-created DBs nor JSONL canonical files, blobs, or artifacts, despite split-brain migration then archiving the entire realm. A live durable writer can modify the supposedly quiesced losing copy during comparison/archive. Use a stable realm-level admission fence spanning every durable backend and the entire compare-to-rename interval.

4. [P1] Split-brain reporting omits authoritative durable data. compute_split_brain_report compares sessions.sqlite3 and enumerated SQLite files only (meerkat-store/src/migrate.rs:953-1040). It never compares canonical sessions_jsonl/*.jsonl, blobs, or artifacts. A JSONL realm with divergent sessions but equal/absent derived indexes can be reported as non-divergent, after which --adopt-root archives the only copy. The report must cover every authoritative store before it can support an archive decision.

5. [P1] Fresh preservation archives can still be pruned immediately. The archive filename records the current timestamp (meerkat-store/src/migrate.rs:155-165), but fs::rename preserves a long-idle source's mtime and prune still computes age from that mtime (meerkat-store/src/migrate.rs:1094-1100). A realm archived today can therefore already exceed the default 30-day retention threshold. Base age on the registered archive timestamp or record a new archival time.

6. [P1] Long-lived task stores stop checking the schema ledger after their first operation. SqliteTaskStore::open_cached calls apply_domain_migrations only while schema_ensured is false (meerkat-tools/src/builtin/sqlite_store.rs:66-92). After a newer process migrates tools-tasks, an older live handle opens further connections and continues reading/writing without a future-schema refusal. Recheck the ledger on every physical open, or make the cache generation-safe.

7. [P1] storage prune accepts --realm but ignores it. handle_storage_prune builds PruneOptions with roots/apply/age only at meerkat-cli/src/main.rs:7748-7760; doctor and migrate apply a realm filter, prune does not. rkat --realm alpha storage prune --apply can delete eligible artifacts for every realm beneath the roots. Carry and enforce the requested realm before enumeration/deletion.

These are static-control-flow defects in the current head; the root-selection and migration/retention cases can respectively create unrecoverable split realms or delete the only preserved copy.

lukacf:
All seven P1s are addressed, plus the follow-on P2/P3 batch. Head: the four commits ending at `fix(storage): second-review hardening`.

**P1 dispositions:**

1. **Layout in the production open path** — surfaces (CLI incl. mob-host, RPC, REST, MCP server) now resolve one `StorageLayout` (project root walked up from the invocation context) and open via the new `open_realm_persistence_with_layout`, which threads the layout into `RealmOpenContext` and the provider seam.
2. **Dual-root first-start race** — probe metadata errors are typed (`RootProbeFailed` / `ManifestUnreadable`; only NotFound means absent), and first materialization runs a cross-candidate reservation protocol (`create_new` markers in every candidate root, re-probe under reservation, materialize, release; bounded wait on contention, typed `MaterializedElsewhere` refusal, age-based stale-marker takeover). Aliased candidate roots collapse by canonicalization; sanitized-id collisions refuse typed with both identities named.
3. **Realm-wide fence** — `RealmMaintenanceFence` now takes a realm-level write-admission fence plus the full fixed inventory (existing or not) and the re-enumerated `mobs/*` set; JSONL/blob/artifact write paths participate via shared per-realm admission guards (typed `MaintenanceFenceHeld`, acquired before any lock-file mutation), and the CLI fences every twin **before** divergence comparison, holding through the archive renames.
4. **Split-brain report coverage** — the report now covers `sessions_jsonl/*.jsonl`, blob and artifact trees (relative path + content digest); `--adopt-root` refuses while any comparison entry is unknown/unreadable (`comparison_is_conclusive()`), read failures poison the comparison instead of reading as absence, and hashing streams instead of `fs::read`.
5. **Prune aging** — age derives from the registered name's embedded timestamp (`.pre-<version>-<ts>` / `.corrupt-<ts>`), so a renamed long-idle source can no longer be pruned the day it was archived; mtime is only a fallback for the impossible case.
6. **Task-store ledger recheck** — the once-per-process latch is gone; every physical connection open re-runs the domain ledger check, so a live handle refuses typed (`SchemaFromTheFuture`) after a newer binary migrates the file.
7. **Prune `--realm`** — enforced: realm-scoped enumeration only returns the named realm's artifacts (root-level archives match by archived directory identity), with a CLI test proving another realm's eligible artifacts survive a scoped `--apply`.

**P2/P3 batch** — also landed: ledger preflight before WAL conversion on every primary opener (per-store domains), WAL establishment verified, ledger shape/version validation (TEMP shadowing inert, non-positive versions refused), savepoint-based migration transaction custody, clamped busy timeouts, honest read-only write-contact vocabulary, store adoption of the shared error classifier, `ensure_schema` routed through the ledger, runtime legacy migration certified, doctor rebuilt per the list (JsonColumnBytes, all-domain census, manifest-pin diagnosis, wrong-type/unreadable findings, single read snapshot, bounded dedup, deterministic output), migrate/prune honesty items (dry-run errors, partial-archive visibility, hardening warnings, parent fsync, symlink safety, quarantine restricted to files), CLI scope items (explicit roots don't read ambient HOME, `--isolated`/`--default-model` rejected for storage verbs, attached flag forms parsed), ambient-gate holes closed (prefilter superset, import/alias/const tracking, Windows variants, function-scoped allowlists, build.rs coverage), session bulk-adoption failure accumulation + cursor-ambiguity guard, and the conformance harness items (real legacy fixtures, transcript-rewrite + create-conflict chapters, capability identity proofs, reopen-verified durability, blocking sqlite CI lane, wasm feature split).

**Also in this push** (from the production report): the idle busy-loop fix (machine schemas cached once per process — an idle durable member cost ~0.3 core in 0.8.2/0.8.3), a stream-inactivity watchdog (default ON, 300s) so a stalled provider continuation fails typed instead of hanging forever, and `force_cancel` made legal from `Running` (idempotent when already cancelling, `MemberNotFound` for unknown identities).

Verification: workspace unit 9,345/9,345 · integration 11,512/11,512 · e2e-fast 9/9 · clippy `--workspace --all-targets --all-features -D warnings` clean · wasm32 check+clippy clean · machine-check-drift, rmat-audit --strict, effect-authority, bridge-classifier, storage-ambient-gate, machine-authority docs + poster gates all green.

PR #922, opened 2026-07-27
0.8.9: the flat turn curve — witness-v3 + O(delta) turn boundaries

lukacf:
## 0.8.9 is the flat-curves release

An identical one-word turn must cost the same whether the accumulated session document is 256 KB or 10 MB. It now does.

### Measured (gate fixture, identical one-word turns, per-site canonicalized-and-hashed bytes)

| | large member (~9 MB doc) | small member (~256 KB doc) |
|---|---|---|
| before | 435 MB/turn | 6 MB/turn |
| after | **0 MB/turn** | **0 MB/turn** |
| ratio | **1.0× (asserted in-lane, `e2e_smoke_mob_turn_latency_gate`)** | |

Site split after: every bucket (witness, decode, checkpoint-digest, append-guard, boundary-guard, rewrite-chain-walk, other) reads 0 on ordinary turns at both sizes. The one deliberately retained O(document) axis is the single whole-document boundary serialize per commit (15 MB/turn at the large member), pinned by its own per-fixture envelope and named in the CHANGELOG as the next tracked item. Production defect context: live 0.8.6 fleets measured 60 s (14 MB) and 180 s (94 MB) per one-word turn from exactly this class.

### The five attributed drivers, and how each was removed (no check deleted)

1. **History witness, 109 MB/turn → 0 — witness-v3** (`docs-internal/witness-v3-migration.md`, implemented as approved): format-3 witness = domain-separated digest over (head revision, canonical commit log digest, sorted retained revision-id set) — O(retained revisions), never O(retained body bytes); body-byte integrity is owned by seal-at-ingress, where it was already proven. Two version axes (`digest_format` stays 2 untouched); typed full/slim carrier normalizing bare strings to v2; accepted formats registered in the generated `SessionPersistenceVersionAuthority`; **checkpoint stamp schema 3 = max(provenance floor, witness floor) is the downgrade one-way door** (pre-v3 binaries refuse typed via their existing future-schema path); v2 evidence verifies as v2 forever (mixed stores, lazy per-session upgrade on the first authoritative full-graph write); unknown formats refuse at document ingress before any normalization or healing.
2. **Per-decode graph validation, 55 MB/turn → 0 warm**: producer seams admit the graphs they prove into the validated decode memo (substitution semantics; admission requires full validity — FullVerify or a construction-plus-content proof; unprovable shapes fall through to FullVerify). Kill-switch `MEERKAT_DISABLE_GRAPH_DECODE_MEMO` restores full cost end to end (red-first-proven in a child process).
3. **Boundary-guard re-proof, 47 MB/turn → 0**: guards demand the sealed `ValidatedTranscriptHistory`; per-commit whole-graph loops collapse to the membership fact the seal doesn't state. The adoption arm is strictly **strengthened** (chain coherence was never checked there before — new test pins the previously-accepted non-linking chain as rejected).
4. **Compaction-path revalidations, ~6× per compaction → ~1 required rebuild hash**: authority minted from the accumulator digest + one hash of the rebuilt transcript; no-op answered from the token's own digests; whole-span digests reuse the one required hash (partial spans keep paying O(span), pinned); rewrite commits prove their successor graph by construction + an O(1) chain-content proof, with a **fail-closed release-sampled full-validator cross-check**.
5. **Checkpoint canonical passes, 36 MB/turn → 0 warm — framed midstate**: retained SHA-256 midstate over the document's transcript span (`created_at`/`id` sort before `messages` and are immutable), spliced with the O(metadata) suffix at a NUL-prefixed exactly-once marker. **Digest values are byte-identical** — pinned against a committed fixture minted by the published 0.8.8 crate, with sampled full-path cross-checks and fail-open fallback.

Plus the trace-attributed residuals: `replace_synthetic_notices` no longer hashes the document twice per LLM call to answer a no-op question about its own notices; accumulator midstates ride across serialize/decode via a memo keyed by the SHA-256 of the exact serialized bytes (the byte-binding discipline the P0 memo review required); slim materialization serves the head writer's proven vector only on memcmp equality (row tampering still fails `Corrupted`); the live session warms the framed midstate once per lifetime so every per-save clone and decoded copy inherits it.

### Adversarial reviews (two independent lenses, all findings addressed in-tree)

- **Integrity lens**: P0 — the rewrite fast path accepted graphs the full validator rejects (mechanical-head chain-content gap; warm-persisted, memo-laundered, unloadable at restart). Fixed exactly along the recommended line: chain-content proof (empty log | endpoint parent | boundary-ring prefix witness), FullVerify fallback, and the reviewer's red scenario pinned green (`rewrite_on_a_pointer_extended_mechanical_head_fails_closed_and_stays_decodable`). P1 (seeding soundness) resolved by the same fix; both P2s (marker-trust record, memo-doc honesty) done.
- **Byte-identity lens**: **P0 explicitly none — the byte-identity proof holds** (canonical streams reconstructed end to end). P1 (zero tests) resolved by the suites below; P1 (verification topology: no release-build byte verification between rewrite-commit and cold restart) fixed — the fast path now verifies **fail-closed** under the shared release sampling budget, raised 32→64 to cover all three sampled cross-checks. P2s dispositioned in-tree or recorded.

### Tests (all green: meerkat-core 1565/1565; store+session+runtime+mob 4396/4396; e2e-fast 9/9; clippy/wasm clean)

- `meerkat-core/tests/fixtures/` — **real 0.8.8 bytes** (generated by the published crates.io crate, verified in a separate cold process so 0.8.8's own memos could not vouch).
- `witness_v3_migration.rs` (16 tests): v2-verifies-under-v3-binary, schema-3 mint vs graph-free floor, full→slim→full carrier round trip, slim-cannot-relabel, unknown-format ingress refusal, head/commit/revision-set/body-byte mutation proofs, mixed store, lazy-conversion crash safety, body-bytes independence (never claimed independent of revision count).
- `framed_checkpoint_digest.rs` (5): framed==full across shapes, warm bytes 333 B at both 10 and 400 messages (cold 2.6 KB vs 523 KB), splice-marker fallback, **committed 0.8.8 digest literal pin**, replacement reseed.
- `decode_memo_producer_seeding.rs` (5): decode delta exactly 0 (two counters), forged-body substitution pin, key-pinned corruption refusal, kill-switch child-process red-first proof.
- In-crate: adoption-arm chain-coherence rejection, partial-span digest pin, the P0 red test.
- Gate re-arm: fixture comparability (compaction threshold raised — a compaction turn is inherently O(document) once; the production defect was ordinary turns), honesty floor moved to encode bytes, fixed 4 MiB ratio allowance (still far below one document pass).

### Honest limits

- Flatness is a warm-process property: a cold start pays one full validation per session document, once. The gate runs both fixtures in one process — that is the production shape (long-lived runtime), stated here rather than implied.
- A compaction turn keeps one irreducible hash + persist of the rebuilt transcript.
- Debug builds cross-check every fast-path serve in full (the 27× CPU diagnostic in the gate log is scaffolding, not shipped cost); release samples the first 64, fail-closed.
- Whole-document boundary serialization (encode axis) is unchanged by design and tracked separately.

### Breaking

Graph-bearing session documents mint checkpoint stamp `schema_version` 3; pre-0.8.9 binaries refuse them typed (one-way per session, same contract as every `meerkat_schema` ledger bump). Graph-free sessions keep their provenance floor and stay downgrade-readable. Details under `### Breaking` in the CHANGELOG.


lukacf:
Both Codex findings taken, fixed in a04ace133:

**P1 (fast-path live binding)** — confirmed real, same family as the earlier adversarial-review P0. The fast path now additionally requires `d(live) == commit.parent_revision` (O(delta) on warm sessions; digest equality is canonical-content equality, so the validator's span/prefix/suffix relations hold by construction). Unprovable shapes fall through to FullVerify and fail closed — the pre-change behavior for this exact shape. Pinned: `rewrite_on_a_head_divergent_decoded_row_fails_closed`.

**P2 (witness format at install)** — confirmed, and taken one step further than the suggested fix: schema-outranks-stale-carrier is applied at BOTH the install slow path (the reported refusal) and document-evidence resolution (otherwise the same coexistence document, once stamped schema-3, would be refused on every *subsequent* verify — the reported fix alone would have moved the failure rather than removed it). The carried witness is still independently cross-checked under its own format, and slim rows are untouched. Pinned: `schema3_stamp_installs_and_verifies_over_a_stale_carried_v2_witness`.

meerkat-core: 1568/1568 green with both fixes.

PR #977, opened 2026-08-20
release: preflight npm package write access

lukacf:
## Summary

- add a live npm publication preflight immediately after exact-main CI
- require the release token to authenticate and hold `read-write` access to both `@rkat/sdk` and `@rkat/web`
- gate both registry-publishing jobs on the successful preflight so crates.io and PyPI cannot publish before npm failure is known
- preserve the Rust-only alpha lane without requiring npm credentials
- keep the token in a mode-restricted temporary npm config rather than command arguments or logs
- add failure-path/selftests and a release-doctor ratchet for the workflow dependency

## Why

Meerkat 0.8.24 published all 42 Rust crates and the Python SDK before the shared npm token failed at the final package steps. The release therefore became partially published. This change makes npm identity and both package-write permissions an early read-only gate before any registry write begins.

## Follow-up test hardening

- make the queued-input census test hold the accepted input at a deterministic executor barrier before asserting the live ledger row and DSL phase-key correspondence
- replace one-second RPC scheduling guesses with signal-driven witnesses already validated on PR 976
- production runtime behavior is unchanged by these test-only follow-ups

## Validation

- `scripts/test-release-npm-publish-preflight.sh`
- `make release-doctor` - 20 pass, 2 expected local warnings, 0 fail
- `actionlint .github/workflows/release.yml`
- `shellcheck scripts/release-npm-publish-preflight scripts/test-release-npm-publish-preflight.sh`
- exact queued-input census test - 1/1 and 50/50 green
- mandatory detached pre-push hook at `626c2bd8c` - full pass, including changed-crate Clippy, workspace unit and integration tests, cold restart, and e2e-fast
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -381,6 +381,41 @@ jobs:
           echo "BuildBuddy release validation skipped: ${buildbuddy_result}"
           [[ "${cargo_result}" == "success" && "${buildbuddy_result}" == "skipped" ]]
 
+  # Authenticate before any registry job can make an irreversible write. The
+  # same token publishes both npm SDKs, so require live write access to both
+  # package names before Rust crates or the Python SDK begin publishing.
+  registry_credentials_preflight:
+    name: Verify registry publication credentials
+    needs: [require_ci_green]
+    if: >-
+      always() &&
+      needs.require_ci_green.result == 'success' &&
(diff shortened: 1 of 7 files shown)


lukacf:
@codex review

PR #1084, opened 2026-09-03
fix(mob): report explicit resume preparation progress

lukacf:
## Summary

- thread the existing internal lifecycle progress signal through explicit-resume attachment/session preparation
- report member-scoped preparation start/completion with position, total, session, and elapsed time
- preserve the fixed 30-second inactivity budget, public error shape, admission ordering, and exact retirement/retry authority
- warn with the exact member, stage, and inactivity elapsed time before returning the existing typed stall error

## Validation

- focused explicit-resume tests: 5 passed
- `./scripts/repo-cargo check -p meerkat-mob`
- `./scripts/repo-cargo clippy -p meerkat-mob --all-targets -- -D warnings`
- `./scripts/repo-cargo fmt --all -- --check`
- standard pre-push gate, including deterministic unit/integration/e2e lane
- Rust quality gate: passed; review warnings addressed
- Meerkat architecture/Rust zealot: approved with no findings

Mutation proof: removing the production actor-loop preparation completion signal makes the production-path regression fail with `LifecycleOperationProgressStalled`, member `prepare-3`, stage `member_attachment_session_preparation`; restoring it makes the same cumulative preparation test pass.

> [!IMPORTANT]
> **MERGE BLOCKED:** do not merge until HomeCore's preserved generation-119 production clone passes repeated cold boots on a fixed build. That external validation is the only merge block.

The separate MobKit pre-Resume durable-reconciliation projection/concurrency regression is intentionally out of scope for this Meerkat PR.

### DIFF
--- a/meerkat-mob/src/runtime/actor.rs
+++ b/meerkat-mob/src/runtime/actor.rs
@@ -15918,6 +15918,7 @@ impl MobActor {
         &self,
         deadline: Instant,
         admission: super::state::LifecycleAdmissionSignal,
+        progress: &super::state::LifecycleProgressSignal,
     ) -> Result<Vec<ExplicitResumeMemberRebuild>, MobError> {
         let entries = {
             let roster = self.roster.read().await;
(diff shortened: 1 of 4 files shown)


lukacf:
Focused review from the MobKit side (the fleet that hit the 30 s watchdog on 2026-09-03).

Read for two things, both hold:

1. Progress is emitted inside the serial preparation loop, per member, in both the original-session and replacement-session branches: `awaiting_member` before the prepare (sets member/stage/observed_at, no epoch bump) and `member_progress` after it (epoch bump). Since the handle's window re-arms on an epoch change, a 16-member fleet at a few seconds each can no longer sit silent for 30 s between two members.

2. The stall WARN names the member, the stage (`member_attachment_session_preparation`) and the elapsed time since the last observation, so the next fatal attributes itself instead of reading as a bare "made no progress".

One property worth stating in the PR text so nobody reads more into it: a single member whose preparation itself exceeds 30 s still trips the watchdog (the epoch only advances on completion). That is the intended scope here (no timeout change) and it is the right shape now that MobKit #388 removes the 171 s walk; I only want it written down.

No timeout, config, machine, wire or public-error change. Approving.

PR #1085, opened 2026-09-03
fix(mob): scope explicit resume retirement deadlines

lukacf:
## Summary

- keep the public Resume deadline as the stale queued-command admission bound
- give every exact member attachment retirement a fresh existing timeout budget
- let admitted process-owned Resume reach its durable transition under the existing progress authority

## Intended trade

After lifecycle/attachment admission there is intentionally no fleet-wide absolute total bound. Liveness remains bounded by #1084's typed 30s no-progress window across members and the durable transition, while each exact member retirement receives its own fresh typed 30s budget. This prevents healthy sequential progress from exhausting the first caller's deadline without introducing an unbounded opaque await.

## Validation

- focused `explicit_resume` tests: 7 passed
- cumulative public Resume regression: five sequential 550ms preparations exceed the 2s test deadline and still commit Running exactly once
- shared-deadline mutation: killed at a later exact attachment retirement
- retained single-member stall remains `LifecycleOperationProgressStalled`
- `check -p meerkat-mob --all-features --all-targets`
- strict all-target Clippy with `-D warnings`
- format check and full exact-tree pre-push gate
- rust-quality-gate: PASS
- meerkat-rust-zealot: APPROVED
- MobKit lead review: APPROVED

### DIFF
--- a/meerkat-mob/src/runtime/actor.rs
+++ b/meerkat-mob/src/runtime/actor.rs
@@ -15916,7 +15916,6 @@ impl MobActor {
     /// machine journal still carries a stale session binding for it.
     async fn prepare_explicit_resume_member_sessions(
         &self,
-        deadline: Instant,
         admission: super::state::LifecycleAdmissionSignal,
         progress: &super::state::LifecycleProgressSignal,
     ) -> Result<Vec<ExplicitResumeMemberRebuild>, MobError> {
@@ -15987,7 +15986,7 @@ impl MobActor {
                 .provisioner
                 .prepare_member_session_for_explicit_resume(
                     &session_id,
-                    deadline,
+                    Instant::now() + super::provisioner::EXPLICIT_RESUME_RETIRE_TOTAL_TIMEOUT,
                     Some(admission.clone()),
                 )
                 .await?;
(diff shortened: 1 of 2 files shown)


lukacf:
Review from the MobKit side, of head 3dcc9e4d (the head validated on HomeCore's gen-119 clone as run-9, 2/2 PASS paired with MobKit main 2986bf4b).

Read: the shared caller `deadline` parameter is gone; both `prepare_member_session_for_explicit_resume` call sites (original and replacement session) mint a fresh `EXPLICIT_RESUME_RETIRE_TOTAL_TIMEOUT` per member; the pre-transition absolute check that produced "admission is still pending at lifecycle_transition_admission" and its pre-admission rollback are removed, since after admission the process owns the Resume. No constant changed (30 s prod, 2 s test). Tests cover: admitted Resume outlives the initial caller deadline; each member owns a fresh budget; the production shape (cumulative preparation beyond the initial deadline) completes.

One property worth stating in the PR text: after admission there is no absolute total bound on an explicit resume any more. A stall is detected by #1084's 30 s no-progress window and by the per-member 30 s retirement budget, both typed. That is the intended trade for a 17-member fleet and it is right; writing it down saves the next reader a search for the bound.

Approving from MobKit's side (the formal approve button is blocked because the PR and this account are the same GitHub user).

PR #1087, opened 2026-09-04
fix(mob): mob.toml diagnostics, mob prompt de-dup, author guide fixes

lukacf:
## Summary

A mob author who writes `mob.toml` and reads the guides was getting silent drops instead of diagnostics, a system prompt padded with text the provider already had, and docs that did not say how members, peers, and the WorkGraph surface actually behave. A profile-level `system_prompt` key was absorbed by the untagged `ProfileBinding` and the member ran on the default prompt (MobKit's own fixtures carried the inert key); a typo such as `comm = true` under `[profiles.<name>.tools]` produced the same comms=false wiring rejection at spawn as omitting the key, and that rejection named no fix; every `tools.mob` member paid about 16.6 KB of mob tool descriptions twice per request because the facade re-rendered `ToolDef.description` into `# Available Tools`; and the preloaded `workgraph-workflow` skill, the mobs guide, the comms guide and reference, and `docs/api/rpc.mdx` were missing or contradicting the rules members and hosts kept getting wrong. This PR makes `MobDefinition::from_toml` refuse the three keys that reach for the member prompt with a typed error and a hint, warn on every other unknown key while parsing continues (so a host-private key such as HomeCore's `role_summary` keeps booting), stops rendering Mob-provenance tool descriptions into the prompt, extends the wiring error with its remedy, and brings the docs and the skill in line with the runtime. No `meerkat-contracts` wire type changes.

### meerkat-mob: definition diagnostics (`definition.rs`, `profile.rs`, `error.rs`, `validate.rs`, `lib.rs`, `build.rs`)

- `MobDefinition::from_toml` now returns `Result<MobDefinition, MobError>`; TOML syntax errors arrive as the new `MobError::DefinitionParse(toml::de::Error)` (`#[from]`). Every in-workspace caller already used `?` into `Box<dyn Error>` or `expect`; MobKit's non-test callers use anyhow context, `unwrap_or_else` with Display, or `map_err` to String, so none names the old error type.
- New `MobDefinition::parse_toml` reads the text a second time as a `toml::Table` and compares each `[profiles.<name>]` table against the key set its binding declares: `Profile::FIELD_NAMES` for an inline profile, `ProfileBinding::REALM_REF_FIELD_NAMES` (`["realm_profile"]`) for a realm reference, and the `tools` sub-table against `ToolConfig::FIELD_NAMES`. All three constants are kept honest by a derive-probe drift test that captures serde's own field list.
- A key from the closed `UnsupportedProfileKey` list (`system_prompt`, `prompt`, `instructions`; all three reach for the member prompt so the same hint applies) refuses the parse with `MobError::UnsupportedProfileKey { profile, key }`, on inline and realm-reference tables alike. The message says a profile has no `system_prompt`, the member prompt is `profile.skills` resolved against an inline or path `[skills.<id>]` table, and identity-first hosts may set `DurableAgentSpec.additional_instructions` or a customizer's `draft.system_prompt`.
- Every other unknown key is collected per profile (`UnknownProfileKeys`, sorted), returned on `ParsedMobDefinition`, and projected as `DiagnosticCode::UnknownProfileKey` warning diagnostics in the `validate_definition` shape (one per key, location `profiles.<name>.<key>` or `profiles.<name>.tools.<key>`). `from_toml` calls `parse_toml`, emits one `tracing::warn!` per affected profile (fields `mob_id`, `profile`, `keys`), and returns the definition. Parsing continues and the keys are ignored exactly as before.
- The `tools.comms=false` wiring error in `build_agent_config` keeps its `profile '<name>' has tools.comms=false; mob meerkats require comms=true` prefix and adds the remedy: the default is false so omitting the key counts as false; set `comms = true` under `[profiles.<name>.tools]`; to stop peer messaging keep comms on and use `read_only = true` or a per-spawn `tool_access_policy` deny list. The default itself is unchanged.
- Rustdoc: `Profile::peer_description` is peer-facing discovery metadata (the `peers` tool and `mob.peer_added`), never this member's system prompt; `ToolConfig::comms` must be true for every profile that spawns a member. The two exhaustive `MobError` matches in meerkat-mob (spawn_many failure observation in `runtime/handle.rs`, the runtime parity summary in `runtime/tests.rs`) gained arms; both new variants map to the existing generated `DefinitionError` observation kind.

### meerkat facade: mob tool prompt inventory (`meerkat/src/factory.rs`)

- Survey of the existing conventions first: builtins, external tools, schedule and workgraph render full `## name` plus description entries; comms appends a static posture text; exact deferred-catalog dispatchers omit the inventory entirely. There is no names-only convention, so the mob family follows the cheapest one that exists and is mounted without a prompt inventory.
- `render_tool_usage_instructions` now skips every tool whose provenance is `ToolSourceKind::Mob`, whichever dispatcher mounts it: the 19-tool agent-facing surface (~15.8 KB) mounted through the mob dispatcher and the 12-tool `MobOperatorToolDispatcher` family (~0.75 KB) that meerkat-mob composes into `external_tools`. It renders no header when nothing else is left. Non-mob rendering is byte-identical to before. No Config knob was added.

### meerkat-workgraph: preloaded skill (`skills/workgraph-workflow/SKILL.md`, `src/lib.rs`)

- The skill now states: `parent` edges point from the child (`from_id`) to the parent (`to_id`), are added right after `workgraph_create` (which has no parent field) and are permanent; a failed or cancelled child keeps a `require_success` parent waiting unless the parent's join policy says `accept` (parent may proceed) or `propagate` (parent closes with the child's status); `status` must always be passed to `workgraph_close` because the wire default is `completed`, with `failed` for a refuted hypothesis or a fix that did not work and `cancelled` for work dropped without a verdict; `workgraph_list` and `workgraph_snapshot` hide terminal items unless `include_terminal` is true, and `workgraph_ready` only ever returns live items; `labels` matches every listed label; only a `completed` blocker satisfies a `blocks` edge. Tool descriptions and schemas are unchanged. The skill body grows from 3094 to 4629 bytes.

### Docs (`docs/guides/mobs.mdx`, `docs/guides/comms.mdx`, `docs/reference/comms-reference.mdx`, `docs/api/rpc.mdx`, `docs/reference/workgraph-reference.mdx`, `docs/guides/workgraph.mdx`)

- mobs guide: a Warning in "Define A Mob" quoting the exact wiring error and its two surfaces (JSON-RPC -32602 on `mob/spawn` and `mobkit/ensure_member` with the `ensure_member failed: ` prefix; outcome `broken` with the reason in the bootstrap status error on the identity-first plane) and why comms is mandatory; a troubleshooting row next to `profile_not_found`; a new "Member system prompt" subsection (roster blurb vs `profile.skills` plus `[skills.<id>]`, the default-prompt fallback, `additional_instructions` and `draft.system_prompt`, and the refuse/warn contract above); the Anthropic caching snippet in the profile section (nested `provider_tag` form to opt out or force on, the vocabulary, the `cache_ttl` pairing rule, the flat form failing the whole parse, and the 0.8.33 default boundary).
- comms guide and reference: "Peer identity and naming" (`peer_id` UUID is the routing key, the display name is `{mob_id}/{role}/{member}`, a rename re-mints every name, identity-keyed host-realm agent memory is where old names can survive, MobKit identity-first members appear as `mk--` encoded segments) and "Request and response across turns" (queued peer inputs admit at the next run start; a terminal `send_response` with steer is run start plus preemption, never mid-turn; only steer `send_message`/`reply_to_peer`/`send_request` inject into a running turn; `delegate`, a host-side send-and-wait loop, or a fan-out/fan-in flow are the sequencing shapes; no agent-facing await exists). The MessageIntent accordion says it is receiver-side classification and not `send_request` input.
- `docs/api/rpc.mdx`: `### workgraph/<method>` sections for get, list, ready, snapshot, events, goal/status and attention/list in the session/turn style, with the `WorkItemFilter`, `ReadyWorkFilter`, `WorkGraphSnapshotFilter`, `WorkGraphEventFilter`, `GoalStatusRequest` and `AttentionListRequest` fields, `labels` as a conjunction, `include_terminal` default false applied after `statuses`, `limit` 100 default and 1000 max, the append-only rule (close as `cancelled`, link `supersedes`), and the `{kind, id, label?, summary?}` evidence shape. Only prose was added; the method table the alignment gate parses is untouched.
- workgraph reference: "Filter Semantics" table shared by tools, RPC, REST and CLI, plus the append-only rule; the companion-skill paragraph now says members with `tools.workgraph` receive `workgraph-workflow` preloaded. workgraph guide: standard hosts (rkat persistent surfaces, MobKit gateways) auto-issue the realm plus `default` grant; only an embedding Rust host issues its own.

### CHANGELOG

- `### Breaking`: the `from_toml` return type, the two new `MobError` variants, the new `DiagnosticCode` variant, the new public items, and the `system_prompt` refusal declared as a behaviour change for exact-pinned hosts. `### Changed`: the mob prompt inventory, the skill growth, the comms remedy and rustdoc, and the unknown-key warning.

## Feedback items addressed

- `A3-tool-schema-bloat` (step one only): mob tool descriptions are no longer duplicated into the system prompt (~16.6 KB per request for a `tools.mob` member); the 31-tool schema payload itself is a later step and is untouched here.
- `J1-peer-description-prompt`: `peer_description` is documented as roster metadata in rustdoc and the guide; a profile-level `system_prompt` (and `prompt`, `instructions`) is refused with a hint pointing at `profile.skills`, `additional_instructions`, and `draft.system_prompt`; the "Member system prompt" subsection shows the working TOML.
- `R5-comms-mandatory`: the wiring error carries its remedy; the guide has the Warning callout with the exact text and both surfaces, and a troubleshooting row; a `comm` typo under `tools` now warns with its location instead of silently leaving comms off; the default stays false.
- `J5-comms-ergonomics`: peer naming and request/response-across-turns semantics are documented in the guide and the reference; the closed `send_request` intent vocabulary can no longer be mistaken for the MessageIntent table.
- `R8-workgraph-rpc-surface` (steering half plus RPC docs): the preloaded skill states edge direction, explicit close status, `include_terminal`, and match-all `labels`; `docs/api/rpc.mdx` documents every `workgraph/*` method and filter; the reference gains filter semantics and the guide corrects the grant paragraph.
- `J2-prompt-caching-config` (mob.toml snippet only): the profile section shows the nested `provider_tag` opt-out and force-on forms, the vocabulary, and that the flat form fails the whole definition parse.

## Testing

All compiling commands ran behind `/tmp/feedback-lanes/build-slot.py` from the worktree at HEAD `bc1f78fd3`.

- `./scripts/repo-cargo clippy -p meerkat -p meerkat-mob -p meerkat-workgraph --all-targets -- -D warnings`: exit 0, zero warnings (the package-scoped form was green; the `--all-features` fallback was not needed).
- `./scripts/repo-cargo nextest run -p meerkat -p meerkat-mob -p meerkat-workgraph` (the brief's exact form, default profile): exit 100. Run 1 stopped at 2849/3417 by fail-fast on one failure; run 2 with `--no-fail-fast` executed all 3417: 3413 passed, 13 skipped, 4 failed. The four failures are not caused by this branch (analysis below):
  - `meerkat-mob::host_bind_ceremony` `bind_host_reply_peer_id_mismatch_never_commits`, `bind_host_garbled_reply_fail_stops_until_cold_retry_converges`, `recovered_started_rejection_cannot_abort_prior_uncertain_send_or_unblock_destroy`: each panics in well under a second at `meerkat-mob/tests/support/mod.rs:4039` with `unreachable!("actor fail-stop restart requires the test-support feature")`. The three-package form never enables `meerkat-mob/test-support`: the facade does not depend on meerkat-mob at all, and the feature is only requested by meerkat-rpc, meerkat-rest, meerkat-cli, meerkat-mob-mcp, meerkat-mcp-server, meerkat-web-runtime and meerkat-integration-tests, which `cargo int` (`--workspace --tests --profile fast`) and CI pull in. `tests/support/mod.rs` is untouched on this branch, so this is a property of the command form at any tree state, base included; the round-1 worker's first full run at 01:13 shows two of the same three failures before fail-fast cancelled it.
  - `meerkat-mob::cross_host_flows` `overlay_autonomous_rejected_at_dispatch_same_cause_local_and_remote`: panics at `cross_host_flows.rs:566` (`turn_outcome_rows` not empty) in both full-lane runs, while two other lanes' MobKit nextest runs shared the machine. It passes standalone (1/1, 1.1 s) and when its binary runs alone under the same package set (20/20, 30.6 s), and it passed in the round-1 worker's full run. The file was last touched in v0.8.22, is not changed on this branch, and nothing on the branch is on the flow-dispatch or delivery path. The fast profile serializes this binary into the macOS real-comms cohort for exactly this kind of cross-process contention; the default profile does not.
- `./scripts/repo-cargo nextest run -p meerkat -p meerkat-mob -p meerkat-workgraph --features meerkat-mob/test-support --no-fail-fast` (the same lane with the feature the fail-stop tests need): exit 0, 3425 tests run, 3425 passed (7 slow, 1 leaky), 13 skipped, 0 failed, 467.7 s. The 8 extra tests are the feature-gated ones; the round-1 worker's green run had the same 3425 count. The `cross_host_flows` test passed in this run too.
- `./scripts/repo-cargo fmt --all -- --check`: exit 0.
- `make docs-check`: exit 0.
- `make verify-rpc-surface-alignment`: exit 0 (171 methods; docs typed columns match the catalog).
- Diff stat of the branch against `meerkat-contracts/`: empty.
- Em/en dash grep over every added line on the branch and over all seven commit messages: no hits (the only dashes in the changed files are pre-existing lines this branch did not touch).

Tests added: 11 in `meerkat-mob/src/definition.rs` (`from_toml_refuses_profile_system_prompt_and_names_the_remedy`, `from_toml_refuses_prompt_aliases_that_reach_for_the_member_prompt`, `from_toml_refusal_wins_over_unknown_key_warnings`, `parse_toml_warns_on_host_private_profile_keys_and_keeps_parsing` (covers `role_summary`), `from_toml_logs_one_warning_per_profile_with_unknown_keys`, `from_toml_reports_toml_syntax_errors_as_definition_parse`, `parse_toml_reports_no_unknown_keys_for_recognised_tables_and_realm_refs`, `parse_toml_refuses_system_prompt_on_a_realm_reference_table`, `parse_toml_warns_on_extra_keys_in_a_realm_reference_table_and_keeps_the_binding`, `parse_toml_warns_on_unknown_tools_keys_with_the_sub_table_location`, `from_toml_accepts_declared_profile_keys_and_bare_realm_refs`); 4 in `meerkat-mob/src/profile.rs` (`profile_field_names_match_serde_derive`, `tool_config_field_names_match_serde_derive`, `realm_ref_field_names_match_the_serialized_binding`, `unsupported_profile_keys_are_a_closed_list_with_a_prompt_hint`); `test_build_agent_config_fails_when_comms_disabled` in `build.rs` extended to assert the remedy text; 3 in `meerkat/src/factory.rs` (`mounting_mob_tools_leaves_system_prompt_byte_identical_saving_16kb_of_descriptions`, `render_tool_usage_instructions_omits_mob_provenance_tools_and_keeps_the_rest`, `render_tool_usage_instructions_is_empty_when_every_tool_is_mob_provenance`); 6 in `meerkat-workgraph/src/lib.rs` pinning the skill sentences to the live vocabulary (`WorkEdgeKind::Parent`, `WorkStatus` wire names, the `CloseWorkItemRequest` status default observed by deserializing a request without `status`, the `include_terminal` field on a serialized `WorkItemFilter`, the `FailedChildJoinPolicy` wire names).

Mutation checks (recorded in the commit bodies):

- `06cc3f388`: the byte-identical prompt test was run before the fix and failed on the byte-identical assertion (the prompt carried `## delegate` plus its description); green with the fix.
- `c54f2d02c`: with the base `SKILL.md` swapped back in, all five skill tests fail (0 passed, 5 failed); 5/5 with the new text.
- `2cd4420ea`: five independent mutations (refusal branch removed, unknown-key collection removed, per-profile warn removed, remedy-free wiring message restored, `peer_description` dropped from `FIELD_NAMES`): 8 of 71 selected tests went red, exactly the ones each mutation targets; sources restored byte-identical (cmp), 71/71 green.
- `bc1f78fd3`: five independent mutations (realm references skipped again, tools sub-table inspection disabled, `rust_bundles` dropped from `ToolConfig::FIELD_NAMES`, the Mob-provenance filter made inert, the grouped accept/propagate clause restored): 9 of 42 selected tests went red, exactly the targeted ones; sources restored byte-identical, 42/42 green.

Live-provider or subprocess repro runs: none. Every test in this PR is an in-process unit or integration test over the mock/demo clients; the wiring-error surfaces quoted in the docs were taken from the code paths, not from a live gateway run.

## Behaviour changes for exact-pinned hosts

- Unknown profile key ruling (coordinator, 2026-09-03): `system_prompt`, `prompt`, and `instructions` under `[profiles.<name>]` (inline or realm-reference table) are refused at `from_toml` with `MobError::UnsupportedProfileKey` and the hint above; a definition that booted with the inert key now fails to load until the text moves into a `[skills.<id>]` table (MobKit's fixtures are being cleaned in PR K3). Every other unknown key, including HomeCore's host-private `role_summary`, becomes a `validate_definition`-shaped `unknown_profile_key` warning diagnostic from `parse_toml` plus one `tracing::warn!` per profile from `from_toml`, and parsing continues, so HomeCore keeps booting at the next repin. The check runs on the TOML path only; a `MobDefinition` deserialized from JSON is not inspected.
- `MobDefinition::from_toml` returns `Result<MobDefinition, MobError>` instead of `Result<MobDefinition, toml::de::Error>`; `MobError` gains `DefinitionParse` and `UnsupportedProfileKey`; `DiagnosticCode` gains `UnknownProfileKey`. A host that named `toml::de::Error` in a `from_toml` match must adapt; MobKit's callers do not.
- The system prompt of every mob-enabled session shrinks by roughly 16.6 KB, which moves the prompt-cache prefix once after upgrade. The tool definitions the model receives are unchanged.
- The preloaded `workgraph-workflow` skill grows by about 1.5 KB (3094 to 4629 bytes), which moves the prompt-cache prefix of WorkGraph-capable members once after upgrade.
- The `tools.comms=false` wiring error is longer; its historical prefix is unchanged, so hosts that match on the prefix (MobKit's `ensure_member failed: ` wrapper) are unaffected, while a host that compared the full string would see the new suffix.

## Notes for review

- Deviation from the brief, reported by the Task 2 worker: the brief asked for the warning to come "through the existing `validate_definition` seam". `validate_definition(&MobDefinition)` only sees the typed struct, which by construction no longer holds the dropped keys; carrying them on `MobDefinition` would add a field to a type built by struct literal at 74 in-workspace and 36 MobKit sites and would change `PartialEq` for the definition-mismatch comparisons. The diagnostics therefore use the seam's vocabulary (`Diagnostic`, a new `DiagnosticCode` variant, warning severity) and are produced by `parse_toml`; `from_toml` logs them. No Meerkat surface emits `unknown_profile_key` today; a host that surfaces validate diagnostics structurally calls `parse_toml` and merges. The CHANGELOG and the mobs guide say so.
- Round 1 review found and closed two gaps in the first Task 2 cut: realm-reference tables (`realm_profile = "..."` plus `system_prompt`) were never inspected and the earlier test pinned that gap as correct, and the `tools` sub-table was one level too deep for the check. Both are now covered with tests.
- Rejected review items: a merged `load_toml` seam combining parse, inspection and validation was rejected because it would be a new layer over `from_toml`/`parse_toml` rather than a fix at the root (the repo rule is no new layer to patch an edge case); the request to sequence this PR's merge after PR M1 is a coordinator action, not a worker action, and is recorded in the next note instead.
- The caching paragraph in `docs/guides/mobs.mdx` names the version boundary ("From 0.8.33 the default is `automatic` on every backend that supports it ... Releases 0.8.22 through 0.8.32 default to `disabled`"). That statement is only true if PR `fix/provider-request-compat` (M1) lands in the same release as this PR; if M1 slips, the sentence must be edited before release.
- Not changed on purpose: the `ToolConfig.comms` default stays false (three contracts mirrors would have to move with it); no Config knob for the prompt inventory (reviewers rejected it: it would touch meerkat-core config for a behaviour with no user-visible downside); the mob tool schemas themselves (A3 later steps).
- Not carried from the J5 reviewers' optional suggestions because they are outside this PR's owned files or brief: a CHANGELOG backfill for the v0.7.0 intent closure, softening the `send_response` tool description in meerkat-comms, and a runtime test pinning terminal-steer to `StageRunStart`.
- Existing test renamed: `from_toml_accepts_declared_profile_keys_and_skips_realm_refs` is now `from_toml_accepts_declared_profile_keys_and_bare_realm_refs` (its realm reference was already bare; realm refs are no longer skipped) and it also checks the serializer's `tools` keys against `ToolConfig::FIELD_NAMES`.
- The Task 3 commit body cites a 4552-byte skill; round 1 split the accept/propagate clause, so HEAD is 4629 bytes. The CHANGELOG's "about 1.5 KB" growth figure still holds (1535 bytes).
- Lane-form discovery, not fixed here (outside the brief's scope): a package-scoped `nextest run -p meerkat-mob` without `--features meerkat-mob/test-support` cannot pass the three `host_bind_ceremony` fail-stop tests, because `tests/support/mod.rs` reaches an `unreachable!` when the feature is off instead of the binary declaring `required-features`. The repo's `cargo int` lane never hits this because `--workspace` unifies the feature. Worth a `required-features = ["test-support"]` line on that test target in a separate change.


lukacf:
Release note: this PR is deliberately excluded from the Meerkat 0.8.33 emergency train (which carries only #1088 and #1089). It will be considered for the next regular release.

PR #1092, opened 2026-09-04
fix(providers): tool-schema lowering, caching default, quota, config ingress

lukacf:
## Summary

Two external users hit the same wall from different sides: Gemini and OpenAI rejected meerkat's built-in tool schemas (a root-level `not` in `workgraph_claim`, object-variant `oneOf` in the schedule and mob tools, schemars booleans in `council` and `mob_wire`), an exhausted OpenAI key was retried for 90 seconds as if it were a rate limit, and Anthropic prompt caching had silently been off since 0.8.22 while the docs promised it was on. This PR makes what meerkat sends to providers acceptable to their validators, restores the documented caching default, terminalizes exhausted quota, and closes a config-ingress swallow the review found on three product surfaces.

## What changed

**Tool schemas (meerkat-workgraph, meerkat-gemini, meerkat-openai)**
- `workgraph_claim` loses its root-level `not`; the lease exclusivity is stated in both property descriptions and stays enforced by the claim machine.
- Gemini lowering is rebuilt around a positive allowlist of the Schema proto fields: `oneOf` becomes `anyOf`, `allOf` folds into the parent, JSON Schema booleans are lowered, tuple `items` collapse, `enum` members are stringified. The keyword-by-keyword denylist is deleted.
- OpenAI gets one normalizer for all four emission sites (Responses tools, Chat Completions tools, realtime text adapter, live session tools): local `$ref` inlining, root-level `not`/`enum` removal, root-level combinator folding, typed refusal of a non-object root.
- A deterministic integration matrix runs every built-in tool family (workgraph, schedule, mob agent, mob MCP, mob operator local and remote, comms, memory, builtin) through the Gemini and both OpenAI builders. Three live acceptance tests are registered as the `provider-tool-schema-compat` e2e-live suite.

**Quota (meerkat-llm-core, provider clients)**
- New non-retryable `LlmError::QuotaExhausted`. A 429 whose body names exhausted quota (OpenAI `insufficient_quota` and the spend-limit codes, Anthropic `enforced_spend_limit_reached` and HTTP 402, Gemini `quota_exceeded` and daily `QuotaFailure`) fails in one round trip. Ordinary rate limits keep `RateLimited`. The wire projection carries `details.class = "quota_exhausted"`.

**Anthropic prompt caching (meerkat-anthropic, meerkat-core)**
- The default is `automatic` again on the Anthropic API, Vertex and Foundry, derived from backend capability; Bedrock and the Copilot backend stay `disabled`. One authority for the default across every configured backend. `cache_ttl` is filled by `merge_missing_from`. Realm-level `[agent] provider_params`, which parsed and was silently ignored, is now refused at config load with a pointer to per-profile `provider_params`.

**Config ingress on the product surfaces (meerkat-rpc, meerkat-mcp-server, meerkat-rest)**
- The realm-config refusal above used to be swallowed into `Config::default()` on rkat-rpc, rkat-mcp and rkat-rest, booting them on an entirely defaulted head config. All three now fail startup with the typed error. rkat-mcp also refuses to start when the realm config chain does not compose for a parent document.

## Feedback items addressed

- A2: openai_compatible does not lower tool schemas; OpenAI 400 on `workgraph_claim`.
- X1: Gemini 400 INVALID_ARGUMENT on built-in tool schemas.
- J2: prompt caching not reachable and a Claude run costing far more than expected (root cause: default flipped off in 0.8.22 against the docs).
- A5 (sub-finding): missing or exhausted key retried for 90 seconds before failing.

## Testing

- `clippy --all-features --all-targets -D warnings` on every touched crate and the integration crate: clean. The package-scoped form without `--all-features` is red only on meerkat-runtime dead code identical on main.
- nextest for meerkat-gemini, meerkat-openai (realtime feature), meerkat-anthropic, meerkat-workgraph, meerkat-llm-core, meerkat-core, meerkat-rpc, meerkat-mcp-server: 3608 passed, 0 failed, 25 ignored.
- Integration matrix (deterministic): pass. Live acceptance with real keys: gemini-3.8-flash, gpt-5.5 over Chat Completions and gpt-5.5 over Responses each accepted all nine tool families (workgraph 15 tools, schedule 8, mob-agent 18, mob-mcp 16, mob-operator 12, mob-operator-remote 12, comms 5, memory 1, builtin 11). Before the boolean lowering the same Gemini run rejected `council` and `mob_wire`.
- Fifteen mutation checks, each fix reverted in isolation and the named test observed red, then restored byte-exact (details in the commit bodies).
- `fmt --check`, `make docs-check`, `make verify-lock-consistency`, Bazel BUILD freshness: pass. Cargo.lock unchanged.
- Pre-push hook lane: passed clippy and every governance and contract gate and the full unit lane; one integration test in an untouched crate (`meerkat-skills` stdio source timeout) flaked under load and passes standalone.

## Behaviour changes for exact-pinned hosts

- Anthropic sessions that never set `cache_control` pay five-minute cache writes and read cache hits again (visible in `Usage.cache_creation_tokens` / `cache_read_tokens`). Opt out per profile with `provider_params = { provider_tag = { provider = "anthropic", cache_control = "disabled" } }`; a flat `cache_control` key fails the definition parse.
- `LlmError` gained `QuotaExhausted`; exhaustive matches need an arm (declared under Breaking in the CHANGELOG). Exhausted quota now surfaces as `llm_provider_error` within one round trip instead of `llm_rate_limited` after retries.
- A head `.rkat/config.toml` carrying `[agent] provider_params` is refused, and rkat-rpc, rkat-mcp and rkat-rest exit at startup instead of booting on defaults.
- `workgraph_claim`'s schema no longer carries a root-level `not`.

## Notes for review

- The OpenAI root-combinator fold runs on the Responses path too, so both OpenAI sites share one normalizer even though Responses tolerates root combinators today.
- Gemini `title` stays out of the allowlist (the prior lowering stripped it and a test pins that); `format` values pass through untouched pending a live 400.
- Follow-ups deliberately not in this PR: a dedicated `LlmProviderErrorKind::QuotaExhausted` wire kind (needs `make regen-schemas`; until then `details.class` is the discriminator), and feature-gating the three request builders made `pub` + `#[doc(hidden)]` for the matrix.
- `GeminiClient::build_request_body`, `OpenAiClient::build_request_body` and `OpenAiCompatibleClient::build_chat_completions_body` are public for the integration matrix only.


lukacf:
**Landed on main via #1087's squash commit ba6644317**, not via this PR.

Sequence: this PR's head 5e7c5ba4d had a fully green fresh CI run (33900710243), but the pull_request-event run 33885115276 had failed only on the 20-minute push-to-terminal SLO gate (every job green), and the repository's required-check rule still counted that failed `CI gate` check, so `gh pr merge` was refused. #1087 was stacked on this branch (its history contains all seven commits here) and its head was fully green, so its squash merge carried this PR's content. `git diff 5e7c5ba4d^{tree} ba6644317^{tree}` differs only by #1087's own changes; the tree of main equals #1087's tip exactly.

All seven commit messages of this PR are preserved in the squash body of ba6644317 (repo squash setting = concatenated commit messages). Closing this PR as merged-by-#1087.

Issue #1093, opened 2026-09-04
Stale live-session unregister can wait minutes for runtime-loop handoff under load (durable_jobs_workgraph_recovery FLAKY in Turbo S 33864959013)


lukacf:
Observed in the countable Turbo S run https://github.com/lukacf/meerkat/actions/runs/33864959013 on exact main 239bc2e16 (BuildBuddy invocation 4c8f778e-51aa-47e5-a7da-e19e00bdd9cd): 63/63 shards pass, but `//:e2e_smoke_turbo_s_durable_jobs_workgraph_recovery` was FLAKY (failed attempt 1 after 272s, passed attempt 2).

Attempt 1 did not fail with `UnregisterInProgress` (the pre-#1089 failure). The test client timed out: `timed out waiting for stdio JSON line after 179s`. The reopened `rkat-rpc` stderr shows:

- 10:50:29 realm storage resolved; 10:51:49 ops lifecycle created (80s gap already, aggregate RBE load at cohort start)
- 10:51:51 `unregister_session_inner_locked_authorized start` -> `awaiting runtime-loop and comms-drain quiescence`
- 10:51:56 .. 10:54:46: 35 consecutive `unregister is still waiting for the runtime loop to hand off the exact executor` reports (5s cadence), i.e. the stale incarnation's runtime loop did not hand off for at least 175s
- 10:53:51 `run_progress: staged run has not returned and its execution start is unobservable; the execution-start bound cannot arm for this run`

So after #1089 the `turn/start` caller correctly waits for the exact teardown instead of erroring at 2s, but the teardown itself is gated on a runtime loop that under load takes minutes to hand off. Questions to answer:

1. Why does the stale incarnation's runtime loop not hand off promptly on unregister? Is the interrupt delivered (`UNREGISTER_INTERRUPT_DELIVERY_GRACE`), and is the loop parked in a non-cooperative await (the staged run whose execution start is unobservable)?
2. Should the reopened recovery path avoid materializing a live session that is immediately detected stale, so no teardown is needed at all?
3. Whether an outer bound on the facade wait (see #1090) should exist at all: any bound shorter than the handoff latency reintroduces the error; the real fix is faster handoff.

Related: #1088, #1089, #1090.

lukacf:
**Root cause (read-only code trace, matches every logged line of attempt 1):**

1. The 80s gap after "resolved realm storage" is the test's lease wait (`smoke_shared_realm.rs` ~4064), not load.
[...]

PR #1098, opened 2026-09-04
fix(session): keep runtime registration when no live actor is stale

### LINKED ISSUE #1090: Follow-up to #1089: bound, document, and unit-test stale live-session discard
Non-blocking review findings from PR #1089 (fix(session): await stale runtime teardown), deferred so the 0.8.33 emergency-train head stayed exact.

1. `discard_stale_live_session` (`meerkat/src/session_runtime/runtime_state.rs`) awaits `unregister_session_registration_until_terminal_if_current` with no outer bound in an RPC request path (`turn/start`, input ingress, external event injection at `meerkat-rpc/src/session_runtime.rs:2678/7007/7471`). Consider a 30s outer `tokio::time::timeout` like `meerkat-cli/src/main.rs:632-670`, surfacing a typed error on elapse. Note: any bound short enough to matter must still be longer than RBE teardown latency, or it reintroduces the `durable_jobs_workgraph_recovery` failure.
2. Doc comment on `unregister_session_registration_until_terminal_if_current` (`meerkat-runtime/src/meerkat_machine/session_management.rs`) says "for process-lifecycle owners"; it now also serves exact-registration reoccupation callers. Reword.
3. Doc comment on `discard_stale_live_session` should state the new contract: awaits the exact current registration's teardown to terminal completion; an absent or already-replaced registration is already clean.
4. `.map(|_| ())` drops the `Ok(false)` not-current result silently; add a `tracing::debug!`.
5. Add a focused unit test: (a) absent witness -> `Ok(())`; (b) teardown exceeding the old 2s caller grace completes instead of returning `UnregisterInProgress`.

lukacf:
Refs #1093. Root cause of the durable_jobs_workgraph_recovery flake: in a
reopened rkat-rpc, `monitors/start` calls `ensure_runtime_executor` to
reach the ops registry, which registers a runtime loop + executor with no
live actor. `live_session_is_stale` maps an absent live projection with a
persisted record to "stale", so `discard_stale_live_session` unregistered
that healthy registration and then awaited its owned teardown, which was
gated on a runtime loop parked inside a recovery continuation turn whose
hard-cancel was lost because no actor existed when the interrupt landed.

`discard_stale_live_session` now probes `live_session_actor_registered`
(mechanical registry check, no actor RPC) first. With no live actor it
returns `Ok(())` and leaves the registration untouched, with a
`tracing::debug!`; the callers proceed to `ensure_runtime_executor` +
runtime accept and the in-loop recovery tail materializes the actor from
durable authority, so the new input queues behind any in-flight
continuation. With a live actor the existing path is unchanged: discard
the actor, then await the exact registration's owned teardown to terminal
completion. The error-combination arms are unchanged.

Tests:
- facade: registration with an executor and no live actor keeps the same
  epoch across the discard, the handoff-wait observation counter stays 0,
  and the gated cleanup is never invoked; a live actor still discards and
  unregisters, and a teardown outliving the 2-second caller grace still
  completes with `Ok(())`.
- rpc: reopened-process shape (persisted session, `ensure_runtime_executor`
  as monitors/start does, no live actor) then `start_turn_via_runtime`
  succeeds, the registration epoch survives, the counter stays 0, and the
  recovery tail materializes the actor in place.

Both new keep-registration tests were confirmed to fail with the early
return disabled.

Also carries the #1090 follow-up (doc contract, debug trace on the not-current branch, facade tests for the past-grace wait). Root cause analysis: #1093. Closes #1090. Refs #1093.

Verification: fmt clean; `clippy -p meerkat -p meerkat-runtime -p meerkat-rpc --all-features --all-targets -D warnings` clean; `session_runtime_runtime_state` 10/10; new RPC test `turn_start_keeps_registered_runtime_that_has_no_live_actor` passes and fails with the early return disabled; 35/35 pre-existing stale/recover/ingress RPC tests pass. Targeted RBE `//:e2e_smoke_turbo_s_durable_jobs_workgraph_recovery` result posted as a comment. Pushed with hooks on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -440,6 +440,16 @@ them.
   `MEERKAT_PRE_PUSH_LANE_IDLE_SECS` (default 6 hours). Every decision is
   logged as one `kept:` or `pruned:` line with its reason. Exact-tree
   evidence reuse is unchanged.
+- `discard_stale_live_session` now documents its teardown contract (exact
+  registration witness, terminal-completion wait, absent or replaced
+  registration treated as already clean, preserved error combination) and
+  logs at debug level when the captured registration was replaced before
+  unregister instead of dropping that result silently. The
+  `unregister_session_registration_until_terminal_if_current` doc now covers
+  exact-registration reoccupation callers, not only process-lifecycle owners.
+  Facade tests pin that an absent registration returns `Ok(())` and that a
+  teardown outliving the ordinary 2-second caller grace completes instead of
+  surfacing `UnregisterInProgress`.
 
 ## [0.8.33] - 2026-09-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -460,6 +460,17 @@ them.
   `turn/start`, that an absent registration returns `Ok(())`, and that a live
   actor's teardown outliving the ordinary 2-second caller grace completes
   instead of surfacing `UnregisterInProgress`.
+- The runtime loop no longer retires a registration on a no-pending-boundary
+  terminal while other admitted input is still queued behind it. Previously a
+  detached-job completion wake applied through the in-loop recovery tail with
+  no prompt produced `NoPendingBoundary`, the loop staged
+  `BeginUnregisterSession` and exited, and every queued input (including the
+  `turn/start` prompt admitted moments earlier) resolved
+  `RuntimeTerminated: runtime session unregistered` (refs #1093). The queue is
+  now observed under the driver lock before the terminal is classified, and
+  the teardown prefix re-checks it under the session mutation gate so input
+  admitted in between keeps the registration serving; with nothing queued the
+  no-pending terminal still tears down exactly as before.
 
 ## [0.8.33] - 2026-09-04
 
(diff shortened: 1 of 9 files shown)


lukacf:
DO NOT MERGE yet: the targeted RBE `//:e2e_smoke_turbo_s_durable_jobs_workgraph_recovery` fails on head edb371157 (turn/start -> `runtime terminated: runtime session unregistered`). Root cause is the runtime loop's NoPendingBoundary teardown firing with queued inputs; see the analysis on #1093. A follow-up commit with the loop guard and its tests is being added to this branch.

lukacf:
**Hold lifted pending CI.** Head now carries three commits: the #1090 docs/tests, fix A (keep the registration when no live actor is stale), and the runtime-loop teardown guard for the NoPendingBoundary case (`teardown_after_commit` only when no other input is queued, plus a re-check under the machine gate in `begin_unregister_from_runtime_loop_teardown_unless_queued`). Full analysis on #1093.

Targeted RBE `//:e2e_smoke_turbo_s_durable_jobs_workgraph_recovery` on the loop-guard head 0bcfa2e7f: **pass, first attempt** (invocation c01ea79b-17a6-4ff4-93fc-53ad00cd3c8c). The previous head edb371157 failed this test 2/2, and the new RPC test `turn_start_keeps_registered_runtime_that_has_no_live_actor` reproduces that failure signature deterministically without the guard.

The branch is being re-pushed stacked on #1092/#1087/#1095/#1096/#1099 so the changelog merges cleanly in order; the PR will be marked ready when that push lands.

PR #1118, opened 2026-09-16
ci: run the BuildBuddy lanes for real and fail on unimplemented modes

lukacf:
## Summary

CI has not compiled or tested any code on the BuildBuddy lane since 2026-08-28.

`ci.yml` calls the GCP BuildBuddy workflow with `mode: changed-paths`. `scripts/buildbuddy-ci-lane` never implemented that mode: it fell into the skip arm and exited 0, and the gate read the skip as a pass. The job logs of the last three main commits show it plainly, each in milliseconds:

```
SKIP wasm-check for BuildBuddy mode changed-paths
SKIP cargo-all-features-clippy for BuildBuddy mode changed-paths
SKIP test-unit for BuildBuddy mode changed-paths
SKIP integration-fast for BuildBuddy mode changed-paths
```

It surfaced when Release Turbo S, which builds `sdks/web/wasm` for the browser scenarios, failed on a wasm32 break in meerkat-runtime that CI had passed twice (fixed separately in the own-frame wasm32 PR).

## Changes

- `ci.yml` calls the lane with `full-fresh`, the workflow's declared default, so every Bazel lane runs on the remote executors.
- `scripts/buildbuddy-ci-lane` refuses an unimplemented mode with exit 2 instead of reporting an unrun lane as passed. The intentional per-lane skips for `workspace-warm` and `e2e-system` are unchanged.

## Expectations

This is the first real CI run in three weeks. The 1200 s BuildBuddy SLO and the 2400 s push-to-terminal budget may bite, and lanes may surface failures that accumulated unseen. Both outcomes are information, not reasons to keep the lanes off.


lukacf:
## Status after the first real runs (release owner note)

With the lanes actually running, two classes of accumulated failure surfaced. One is fixed in this PR; the other needs remote iteration and is left for a follow-up decision.

1. **Fixed here:** `meerkat_contracts_unit_test` (prebuild lane) could not read the three tool-policy fixtures it `include_bytes!` from `meerkat-core/tests/fixtures`: the generated BUILD declared no `compile_data` for cross-crate includes. The generator now exports and references any `include_str!`/`include_bytes!`/`include!` target owned by another workspace member (commit `build(bazel): export and reference cross-crate include files`); this also covered the runtime tests' migration fixture and the codegen targets' kernel includes.

2. **Open:** `test-feature-matrix-lib` fails in `meerkat/tests/brain_swap_surface_parity.rs` (added 2026-09-01, after the lanes went dark). Its `repo_root()` falls back to the current directory when `MEERKAT_WORKSPACE_ROOT` is unset; under RBE that is the sandboxed execroot, where only declared inputs exist, so reads of `meerkat/src/surface/runtime_backed.rs`, `meerkat-cli/src/main.rs` and `meerkat-web-runtime/src/lib.rs` fail even though several `package_runfiles` are in `data`. The generator needs to point `MEERKAT_WORKSPACE_ROOT` at the runfiles tree (and add `//meerkat-web-runtime:package_runfiles`) for every repo-scanning test. Likely the same for other governance tests written since 2026-08-28. Expect further lanes to surface issues once these two pass.

3. The wasm-check lane fails until #1119 (the wasm32 own-frame fix) is on main; this branch should take main after that merge.

The 1200 s BuildBuddy SLO has not been measured against a real full-fresh run yet.

lukacf:
Superseded by #1156. Pull-request CI is now Cargo-only on GitHub-hosted runners (Bazel was never meant to gate PRs); this branch's release-lane fixes (BUILD generator gaps, TLC JDK_JAVA_OPTIONS, render_contracts sizing, MEERKAT_WORKSPACE_ROOT, sandbox cargo-clippy, SDK suite split, per-member lint table, unique test names, Python SDK reset, canary linker policy, nested-child fixes, fail-on-unimplemented-mode) are cherry-picked into #1156, where the full BuildBuddy graph runs in nightly and on the release tag. Its ci.yml mode and budget changes are not carried over.

PR #1120, opened 2026-09-16
fix(live): retire an unreported playback target when new provider output supersedes it

lukacf:
## Summary

Public Live aborted the session task when the provider started its next response before the caller had reported playback of the previous one: a barge-in answer, or a follow-up the model volunteers. Release Turbo S reproduced it on both remote attempts of scenario 97 (`assistant_playback_target_already_active`, panic on the machine cleanup thread, then a dead session), and it reproduces locally about one run in three.

## Cause

The transcript reducer keeps one active playback target per session. The generated session-document authority admits the new output, but the reducer rejected the second admission while the first target had no terminal, and a reducer rejection at runtime is treated as a fail-closed durable-restore failure (panic). The window opened with #1117: earlier, the quiet-period heuristic finished turns before the next one started.

## Change

- At admission, an active target for a different item with no pending terminal is retired as `Unmeasured` (no playback evidence was ever reported) before the new target is admitted. If the previous target already has a pending terminal awaiting its provider final, admission fails with a typed error instead of a panic; that ordering cannot occur on the public path, where the broker finishes the open turn before the next one starts.
- A caller report that lands after such a retirement (the client's `live/playback_complete` or `live/truncate` for the superseded output) replays the `Unmeasured` settlement and commits nothing, so the client gets the committed disposition rather than `playback retry conflicts with the committed settlement evidence`. Supersession retirements are distinguished from a caller's own `Unmeasured` truncate by carrying no completion; conflicting reports against caller-made settlements are still rejected (existing test unchanged).
- Unit test `new_provider_output_retires_an_unreported_target_and_accepts_the_late_report`.

## Verification

- meerkat-session authority suite and facade/rpc live suites: pass.
- Real API, scenario 97 and 98 runs: see the run list in the final comment on this PR.

Not addressed here: scenario 97's greeting phase can still time out when the provider keeps interrupting itself during the armed barge-in fixture (provider-side turn cycling, no meerkat error in the log). That is test robustness, tracked separately.


lukacf:
Real-API runs on this head (x86_64 Linux, gpt-live-1): scenario 97 passed 3 of 6 runs; the 3 failures are all the greeting-phase provider pattern timeout with no meerkat error (no panics, no settlement conflicts). Scenario 98 passed 2 of 2. Baseline on the released v0.8.38: scenario 97 panicked with the same reducer rejection in 2 of 3 runs, so the crash class predates #1117; the greeting-phase flake could not be measured there because the runs panicked first.

PR #1125, opened 2026-09-17
feat(live): async voice context, provenance repair, and release v0.8.40

lukacf:
## Summary

Ships the async voice context work as Meerkat 0.8.40, taking over the emergency WIP left on this branch (handover at 58ba9d92b) and finishing it.

What lands on top of 0.8.39:

- **Concurrent bootstrap** (`LiveContextBootstrapMode::Concurrent`, opt-in): live media opens while a bounded summary of the historical conversation is generated; `LiveContextPreparationStatus` exposes the preparation lifecycle independently of media.
- **Observation provenance**: the generated machine admits one ordinal per provider turn and freezes the exact summary-ACK cut. Only rows admitted before the cut are reasserted after the summary; fresh post-ACK output is never echoed back. This closes the self-echo the 73a0b869 baseline had (69 thinking appends in the failed paid run).
- **Summary delivery lane, measured against the real provider.** The bootstrap summary travels on the provider's instructions lane, framed as conversation history that anything said during the call supersedes and that is not to be answered or acknowledged aloud. Every real-API run with the summary on the thinking lane ended with the model answering "I don't know yet" to a recall question asked seconds after the acknowledged summary; the provider does not treat that lane as recallable knowledge and acknowledges it only at a turn boundary, which also left a pending append blocking close for as long as we measured (60 s).
- **The causal tail stays quiet and load-bearing.** Rows committed between the summary snapshot and its acknowledgement are replayed on the thinking lane after the summary so the model keeps the live order of facts; speech after the acknowledgement is never re-sent. Measured: every run that answered the current-facts question correctly had that replay in flight (its content is injected promptly even when the acknowledgement waits for a turn boundary), and every run without it answered with the older typed value. Replaying the same rows on the instructions lane steered the model into delegating, so the thinking lane is the right carrier.
- **Spoken owner context waits for the user.** Typed rows voiced as commentary and delegation results are never started while the provider reports an open user turn (bounded at eight seconds). In one run the drained typed-correction commentary was spoken over the user's next question and the provider never transcribed it. Typed rows stay on the commentary lane on purpose: delivering them as quiet knowledge made an older typed correction the newest user fact and the model answered with it, twice; voiced as commentary the model kept the user's later speech authoritative in every passing run.
- **Close cannot stay stuck.** Measured: with a quiet append pending, the provider withheld `session.closed` for the full sixty seconds we waited, and muting input first did not change that. Close now gives the provider one second in that state and then retires the transport locally, logged as unconfirmed closure; a `session.close` accepted but never confirmed with nothing pending is retired after twenty seconds, with earlier attempts retained for retry.
- Host-designated conversational human input, provider-initiated assistant output admission, close custody revocation before physical retirement, WASM feature gating.

Fixes applied to the WIP in this PR: the production agent forwards provenance-aware playback target admission (previously every observed assistant target failed closed), admission failure on a closed or revoked channel degrades to an unsequenced row instead of ending the provider stream, one ordinal per playback segment instead of per transcript delta, wire and payload-size conversions for the new transcript variants, the instructions lane for the summary, and the user-turn gate.

Decided against and not deferred: a statistical pass rate for the smoke lane (the retry plus the evidence journal's failure classification is the tool), automatic native-seed reopen when a summary fails (the typed failure status is the honest surface), and pruning the per-id observation maps on channel close (growth is one small entry per playback segment).

## Verification

- `check --workspace --all-features --all-targets`, `clippy --workspace --all-features --all-targets -D warnings`, fmt: clean.
- machine-check-drift, schema/SDK codegen freshness, version parity, machine-authority docs, runtime-authority-bypass, poster coverage: pass.
- Focused nextest on core, session, runtime, live, openai, contracts, facade, rpc, machine-schema, machine-kernels: pass except `meerkat-rpc` `router::tests::admitted_remote_answer_uses_injected_strategy_and_machine_result_projection` and `webrtc_only_composition_installs_transport_neutral_lifecycle_cleanup`, which fail identically on released main under `--all-features` (pre-existing, not touched here).
- Offline regressions: generated-authority ordinal/ACK-cut suite; `observation_cut_survives_deferred_materialization_and_late_resolution` extended with a fresh post-cut assistant row.
- `make semver-breaks` against the published 0.8.39 baseline: 3061 findings, all named under `### Breaking`.
- Real API: S99 (`e2e_scenario_99_gpt_live_public_concurrent_context`) results are in the final comment; the scenario now also asserts that thinking-append attempts do not grow between the two post-ACK exchanges.

## Review notes

Two independent adversarial reviews were run on the delta; their remaining non-blocking items are listed in the final comment.


lukacf:
## Real-API evidence and review leftovers

**S99 `e2e_scenario_99_gpt_live_public_concurrent_context`** (real gpt-live-1 media through the browser harness, real gpt-5.5 executor, synthetic microphone) on the final tree: **run 25 passed end to end** in 123 s. Stages in order: opening, connected, initial unknown (honest "I don't know yet" before the summary), typed correction, spoken correction, delegated shell work, releasing summary, provider acknowledged at 45 s, historical recall (exact vault phrase), current facts recall (Cobalt and Marigold), close, reopen, close, reopen, obsolete job release, replacement unknown, replacement summary acknowledged, replacement recall, close. Every close completed inside the five second ceiling. The thinking lane carried only pre-acknowledgement rows; no speech after an acknowledgement was queued at any point including the close flush.

Twenty five runs were needed, and every failure was diagnosed from the evidence journal:

| runs | outcome | cause and disposition |
|---|---|---|
| 1, 2, 6, 8 | recall "I don't know yet" | summary on the thinking lane (not recalled); the seed's "without repeating the vault phrase" leaking into summaries; the question's honest-unknown escape. Summary moved to the instructions lane with history framing; seed and fixture corrected |
| 3, 5, 15, 19, 21 | barge-in or delegation of a recall question | provider variance; Turbo S retry tolerates it, the journal classifies it |
| 4, 9, 12 | recall and current facts passed; failed on my count-based echo check, a close ceiling, or the reply window | test fixes: content-based echo check, provider-timeline reply window |
| 7 | 40 reasserts flushed at close | segment ordinal inheritance (fixed: one ordinal per segment start) |
| 10, 13, 22, 23 | reply window polluted by voiced commentary or a trailing punctuation delta | reply delimited by provider start time; the scenario now lets the assistant finish speaking before each post-acknowledgement question |
| 11, 14, 20, 24 | close never confirmed with a quiet append pending (60 s measured; muting input did not help) | close waits one second in that state and retires locally; unconfirmed close with nothing pending retires after twenty seconds |
| 16, 17, 18 | current facts answered with the older typed value | typed rows as quiet knowledge, or no causal-tail replay; both reverted. The quiet replay is load-bearing |
| 25 | passed | |

Evidence journals for all runs are under `target/e2e-live-audio-artifacts/s99/` on the release VM.

**Adversarial review items decided** (two independent reviews of the delta):

- The six per-id observation maps in the generated machine are never removed on channel close. Growth is one entry per playback segment rather than per transcript delta; kept as is, this is not a follow-up.
- A failure to record the ACK cut for a provider-confirmed summary is reported as delivery-ambiguous and triggers channel replacement. A guard mismatch there would be a bug; replacement is the safe recovery, so it stays.
- Nothing is reasserted any more, so claim-less rows lose nothing; they are logged at debug level.
- `meerkat-rpc` `router::tests::admitted_remote_answer_uses_injected_strategy_and_machine_result_projection` and `webrtc_only_composition_installs_transport_neutral_lifecycle_cleanup` fail under `--all-features` on released main as well; untouched here.

PR #1128, opened 2026-09-17
test(e2e): replace scenario 96 with a live mob fork_off vertical; fix fork_off refusing its running caller

lukacf:
## Summary

Two commits.

**Scenario 96 replaced.** The old scenario ran the deprecated experimental GPT Live path with a ChatGPT OAuth bundle held as a GitHub secret. That bundle expired and failed Turbo S on 2bea37ec5; its coverage already lives in the public scenarios 97 to 99. It is removed along with its test binary, the diagnose workflow and the secret requirement in release-turbo-s.yml. The slot now holds `e2e_smoke_s96_mob_fork_off_vertical` (tests/integration/tests/smoke_mob_fork_off.rs, live Anthropic, about 16 s): a parent member loads a ~3.3k-token ledger so the provider caches a ~23.5k-token prefix, then calls the real `fork_off` tool; the child must answer a planted fact from the inherited transcript and its first request must show `cache_read_tokens` on the whole prefix; an external fork of the running parent must be refused; a second fork with an explicit prefix must know fact 1 and deny fact 2 with a transcript of exactly prefix + 2; both children retire and the parent still answers. This turns the "forks re-bill the parent prefix" question into recurring evidence: the child reads 23.6k cached tokens while `cache_inheritance` still reports `unavailable`.

**fork_off could never succeed live.** Writing the scenario exposed it: the tool's source is the caller's own session, which is always mid-turn while a tool runs, and the durable fork owner refused every running source with `SessionError::Busy`. Fix: `DurableSessionForkTarget` gains `source_admission: DurableForkSourceAdmission { Quiescent, CallerTurn }`. `CallerTurn` is passed only by `dispatch_fork_off` through `fork_member_then_run_bounded` (now an explicit argument). The caller's turn already holds the turn-finalization and recovery boundaries until the tool returns, so the fork takes no lock, skips the competing-writer refusal and the live-session discard, and branches at the last committed snapshot (which mid-turn excludes the running turn's user message and in-flight blocks, so the boundary is always complete). External forks (`MobHandle::fork_member`, RPC, console) keep `Quiescent` and are still refused while the member runs. Declared under 0.8.40 Breaking; `make semver-breaks` passes with every finding named.

## Verification

- Scenario 96 live: 3 passes on the handle path, then 2 consecutive passes with the real tool call after the fix (15.98 s, 16.07 s).
- New unit test: a CallerTurn fork succeeds against a real reserved admission while Quiescent returns Busy; meerkat-session fork tests 8/8, mob and mob-mcp fork tests 46/46, e2e_lanes 28/28.
- clippy all-features all-targets on core, session, mob, mob-mcp, integration-tests; machine drift clean; Bazel generator `--check` clean; `make verify-bazel-locks` green (MODULE.bazel.lock regenerated for the new test target).
- Turbo S will run on the merged head before tagging v0.8.40.


lukacf:
Adversarial review of the CallerTurn fork path (read-only pass over the diff) produced three fixes, pushed as 9c8b112ec:

- **Source write escaped the lock-free path.** The caller-turn loader ran the audit-receipt verification that can append a receipt to the source's event log under `FullReconciliation` coverage, while the running turn appends to the same log. It now verifies read-only and fails the fork closed when a receipt is missing; a later guarded edit repairs it.
- **Enum doc contradicted the implementation.** `CallerTurn` said the owner still serializes against finalization and recovery; it takes no lock because the caller's turn holds those boundaries. Doc reworded to match `caller_turn_fork_mutation_guard`.
- **"Excludes the running turn" is not universal.** A compaction checkpoint inside the turn can advance the committed snapshot to a later complete round boundary. Documented on the enum; scenario 96 now bounds the first fork's length instead of asserting an exact row count.

Confirmed fine by the review: `CallerTurn` reachable only from `dispatch_fork_off` where source equals caller by construction; external forks keep `Quiescent`; guard `Option` fields are `None` only in the caller-turn constructor; Breaking bullet names every new item; no deadlock between the blocked parent and the child's bind and start. Known flake surface of the scenario is model adherence to the literal markers, covered by the Turbo S retry.

